Issue #57, opened 2020-11-20
remove hardcoded passwords in samples


dwpage:
https://github.com/sassoftware/viya4-iac-azure/blob/main/examples/sample-input-ppg.tfvars#L27
(and in a few other sample files)

Could we replace this with random_string or https://registry.terraform.io/providers/hashicorp/random/latest/docs/resources/password

Or at least move the password into a comment, and leave the string as an obvious should be substituted value a la 
location = "<azure-location-value>" # e.g., "eastus2"

thpang:
Closing this one. Found this affects subsequent runs of tf with these values being rotated.

PR #61, opened 2020-11-24
#57 - Removed hardcoded passwords in samples.

thpang:


thpang:
Not implementing at this time.

PR #64, opened 2020-12-04
Subnet warning

thpang:
Added 'coalesce' to the subnet_name assignment calls, removing the warning from terraform.


thpang:
Pulled extra changes. Closing.

PR #65, opened 2020-12-04
Added coalesce to subnet_names removing warning.

### LINKED ISSUE #66: Warning: Interpolation-only expressions are deprecated -for subnet_names
Warning: Interpolation-only expressions are deprecated

on main.tf line 134, in module "vnet":
134: "${local.aks_subnet_name}" = ["Microsoft.Sql"],

Terraform 0.11 and earlier required all non-constant expressions to be
provided via interpolation syntax, but this pattern is now deprecated. To
silence this warning, remove the "${ sequence from the start and the }"
sequence from the end of this expression, leaving just the inner expression.

Template interpolation syntax is still used to construct strings from
expressions when the template includes multiple interpolation sequences or a
mixture of literal strings and interpolations. This deprecation applies only
to templates that consist entirely of a single interpolation sequence.

(and one more similar warning elsewhere)

thpang:
Fixes #66

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -131,8 +131,8 @@ module "vnet" {
   subnet_names        = [local.aks_subnet_name, local.misc_subnet_name]
 
   subnet_service_endpoints = {
-    "${local.aks_subnet_name}"  = ["Microsoft.Sql"],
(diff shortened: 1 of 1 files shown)


thpang:
Fixes #66

Issue #75, opened 2020-12-14
Support max_pods in node pools


NormJohnIV:
Add support to the node-pools for setting max pod count.
https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster_node_pool

thpang:
Adding in docs on reason we need to be able to set `max_pods` for dynamic node pools.

https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/azure-subscription-service-limits#azure-kubernetes-service-limits

manoatsas:
Related to #43

manoatsas:
Fixed in #78

Issue #43, opened 2020-11-12
changes needed to support azure network plugin


BarbaraWalters-sas:
I needed the azure network plugin to install and use cilium which provides detailed performance details for communication. In modules/azure_aks/variables.tf I changed the variable "aks_network_plugin" default value to "azure".

I got an error that pod_cidr was incompatible with the azure plugin.
In the same directory, I modified main.tf and commented out the line that sets a value for pod_cidr

I got another error that said too few addresses were available in the subnet.
In the top level directory, I modified main.tf's value for aks_subnet_cidr_block to "192.168.16.0/20"

This last statement made many more ip addresses available than are necessary, but my Viya 4 deployment was successful after these changes.

I'm not recommending these settings in general, just documenting what worked for me.

manoatsas:
Fixed in PR #78

PR #78, opened 2020-12-16
aks advanced network and nodepools config options

manoatsas:
Closing #43 and #75

### DIFF
--- a/docs/CONFIG-VARS.md
+++ b/docs/CONFIG-VARS.md
@@ -104,7 +104,6 @@ Additional node pools can be created separate from the default nodepool. This is
 | min_nodes | Minimum number of nodes for the nodepool | number | Value must be between 0 and 100. Setting min and max node counts the same disables autoscaling |
 | max_nodes | Maximum number of nodes for the nodepool | number | Value must be between 0 and 100. Setting min and max node counts the same disables autoscaling |
 | max_pods | Maximum number of pods per node | number | Default is 110
-| mode |  Nodepool mode | string | System or User
 | node_taints | Taints for the nodepool VMs | list of strings | |
 | node_labels | Labels to add to the nodepool VMs | map | |
 
@@ -119,7 +118,6 @@ The default values for the `node_pools` variable are:
     "min_nodes"             = 0
     "max_nodes"             = 5
     "max_pods"              = 110
-    "mode"                  = "User"
     "node_taints"           = ["workload.sas.com/class=cas:NoSchedule"]
     "node_labels" = {
       "workload.sas.com/class" = "cas"
@@ -131,7 +129,6 @@ The default values for the `node_pools` variable are:
     "min_nodes"             = 0
     "max_nodes"             = 5
     "max_pods"              = 110
-    "mode"                  = "User"
     "node_taints"           = ["workload.sas.com/class=compute:NoSchedule"]
     "node_labels" = {
       "workload.sas.com/class"        = "compute"
@@ -144,7 +141,6 @@ The default values for the `node_pools` variable are:
     "min_nodes"             = 0
     "max_nodes"             = 5
     "max_pods"              = 110
-    "mode"                  = "User"
     "node_taints"           = ["workload.sas.com/class=connect:NoSchedule"]
     "node_labels" = {
       "workload.sas.com/class"        = "connect"
@@ -157,7 +153,6 @@ The default values for the `node_pools` variable are:
     "min_nodes"             = 0
     "max_nodes"             = 5
     "max_pods"              = 110
-    "mode"                  = "User"
     "node_taints"           = ["workload.sas.com/class=stateless:NoSchedule"]
     "node_labels" = {
       "workload.sas.com/class" = "stateless"
@@ -169,7 +164,6 @@ The default values for the `node_pools` variable are:
     "min_nodes"             = 0
     "max_nodes"             = 3
     "max_pods"              = 110
-    "mode"                  = "User"
     "node_taints"           = ["workload.sas.com/class=stateful:NoSchedule"]
     "node_labels" = {
       "workload.sas.com/class" = "stateful"
--- a/docs/CONFIG-VARS.md
+++ b/docs/CONFIG-VARS.md
@@ -102,9 +102,6 @@ Additional node pools can be created separate from the default nodepool. This is
 | max_nodes | Maximum number of nodes for the nodepool | number | Value must be between 0 and 100. Setting min and max node counts the same disables autoscaling |
 | max_pods | Maximum number of pods per node | number | Default is 110
 | mode |  Nodepool mode | string | System or User
-| enable_node_public_ip | Enable public IP for each node | bool | |
-| priority | VMs priority | string | allowed values - Regular or Spot |
-| eviction_policy | Eviction policy of VMs in Node pool | string | can only be configured when priority is set to Spot, allowed values - Deallocate and Delete
 | node_taints | Taints for the nodepool VMs | list of strings | |
 | node_labels | Labels to add to the nodepool VMs | map | |
 
--- a/docs/CONFIG-VARS.md
+++ b/docs/CONFIG-VARS.md
@@ -100,6 +100,11 @@ Additional node pools can be created separate from the default nodepool. This is
 | os_disk_size | Disk size for nodepool VMs in GB | number | |
 | min_nodes | Minimum number of nodes for the nodepool | number | Value must be between 0 and 100. Setting min and max node counts the same disables autoscaling |
 | max_nodes | Maximum number of nodes for the nodepool | number | Value must be between 0 and 100. Setting min and max node counts the same disables autoscaling |
+| max_pods | Maximum number of pods per node | number | Default is 110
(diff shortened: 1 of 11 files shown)


main.tf
```diff
@@ -74,7 +74,7 @@ data "tls_public_key" "public_key" {
 locals {
   # Network ip ranges
   vnet_cidr_block          = "192.168.0.0/16"
-  aks_subnet_cidr_block    = "192.168.1.0/24"
+  aks_subnet_cidr_block    = "192.168.16.0/20"
```
thpang, line 77: This value should be adjusted based on the user's selection. If the network is `kubenet` then this should keep its old value. However, if it's `azure` then it should be assigned this value. Also what is driving this actual CIDR range: 192.168.16.0./20 why not 192.168.4.0/20 Just wondering why the large leap? This does not collide with any of the previous /24 items.

docs/CONFIG-VARS.md
```diff
@@ -100,6 +100,11 @@ Additional node pools can be created separate from the default nodepool. This is
 | os_disk_size | Disk size for nodepool VMs in GB | number | |
 | min_nodes | Minimum number of nodes for the nodepool | number | Value must be between 0 and 100. Setting min and max node counts the same disables autoscaling |
 | max_nodes | Maximum number of nodes for the nodepool | number | Value must be between 0 and 100. Setting min and max node counts the same disables autoscaling |
+| max_pods | Maximum number of pods per node | number | Default is 110
+| mode |  Nodepool mode | string | System or User
+| enable_node_public_ip | Enable public IP for each node | bool | |
```
thpang, line 105: So not sure why all of these items are being surfaced now: `mode`, `enable_node_public_ip`, `priority`, and `eviction_policy`. I am thinking most of these would be nice to have, but these are not part of this pr. If we need them for this change, please add more details in the pr description. If they are not needed, then they should be removed and placed in another pr.

modules/aks_node_pool/main.tf
```diff
@@ -9,12 +9,21 @@ resource "azurerm_kubernetes_cluster_node_pool" "autoscale_node_pool" {
   proximity_placement_group_id = var.proximity_placement_group_id == "" ? null : var.proximity_placement_group_id
   vm_size                      = var.machine_type
   os_disk_size_gb              = var.os_disk_size
+  # TODO: enable after azurerm v2.37.0
```
thpang, line 12: Azurerm 2.38.0 has been out for about a month now. Is the need for `os_disk_type` tied to this if so lets make the update.

modules/aks_node_pool/main.tf
```diff
@@ -31,11 +40,20 @@ resource "azurerm_kubernetes_cluster_node_pool" "static_node_pool" {
...
```
thpang, line 43: Same comment as above.

modules/aks_node_pool/variables.tf
```diff
@@ -25,6 +25,25 @@ variable "os_disk_size" {
   default = 100
 }
 
+# TODO: enable after azurerm v2.37.0
```
thpang, line 28: Same comment as above.

modules/azure_aks/main.tf
```diff
@@ -1,4 +1,4 @@
-# Reference: https://www.terraform.io/docs/providers/azurerm/r/kubernetes_cluster.html
+# Reference:
```
thpang, line 1: Updated link: https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster

thpang (changes requested):
Looks good overall for the advanced networking, would like some logic changes and some of the changes look like they really need to be in another pr to address the added flexibility of those items and more documentation.

main.tf
enderm: @thpang just wondering why we would want to have a smaller subnet for "kubenet". We have the whole VPC at our disposal, and we are using only a tiny fraction of it. If anything, we should make the subnets larger to make use of the available ip range space so users don't run into issues if they happen to have a larger number of nodes, say.

docs/CONFIG-VARS.md
enderm: does enable_node_public_ip actually make sense? The nodes sit on a private subnet, so there is no way to route to their public ip's directly anyway.

variables.tf
```diff
@@ -102,6 +97,52 @@ variable "default_nodepool_availability_zones" {
...
+variable "aks_service_cidr" {
+  description = "The Network Range used by the Kubernetes service. Changing this forces a new resource to be created."
+  default     = "10.0.0.0/16"
+}
+
+variable "node_vm_admin" {
+  description = "OS Admin User for VMs of AKS Cluster nodes"
+  default     = "azureuser"
+}
+
```
enderm, line 145: I'm not sure about the use case for surfacing the ip and cidr ranges as variables. They could also just be local variables.
I'm thinking about UI simplicity vs functionality.

main.tf
thpang: I can see your point. Currently the total hosts that can be on the AKS subnet is 256. I don't disagree there is always room for growth. With the new CIDR range the AKS subnet will have access to 4096 hosts. If we're going to adjust this value here we should also adjust the same values in the other providers. My suggestion would be to go with: 192.168.4.0/20 to keep things aligned. I am guessing we just took the 192.168.16.0/20 block from Barb's work.

docs/CONFIG-VARS.md
thpang: I was thinking the same thing, but again, believe these value additions should be in another pr.

main.tf
manoatsas: Using the CIDR range based on Azure error message. To @enderm comment it could be opened up to larger range.

enderm: yeah, cidr range calculations are so not intuitive. I found this site helpful for avoiding range overlaps: http://www.davidc.net/sites/default/subnets/subnets.html

docs/CONFIG-VARS.md
manoatsas: These are to expose all available configs to the user, should they want to exercise any and all of these - https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster_node_pool. I could see cloud cost conscious user's wanting to take advantage of 'Spot' VM prices.

Have to check on 'enable_node_public_ip=true' with subnet config, could be removed.

modules/aks_node_pool/main.tf
manoatsas: limiting the scope here to these changes not knowing all the changes that v2.38.0 or the latest v2.40.0 would require.

modules/aks_node_pool/main.tf
manoatsas: same as above

modules/azure_aks/main.tf
manoatsas: Redirects to the link

NormJohnIV:
Since the system/default nodepool is already separate from the additional node_pool what is the reasoning behind allowing setting of the mode in the additional nodepools? They should always be User correct? This seems like it may be an unnecessary option.

modules/azure_aks/main.tf
thpang: What is `Redirects to the link` ? There is no link to reference and/or redirect too. That's why I supplied the link. It was removed from the code.

main.tf
thpang: Ok, then best to keep the 192.168.16.0/20 block. Great tool @enderm Very helpful. Also might need text on this in the code and make sure we propagate this into the other providers.

docs/CONFIG-VARS.md
thpang: I agree, I just don't believe it's part of this pr. They might be needed and do provide flexibility, just not answer this pr's question. My vote would be to open another pr and present these changes there.

manoatsas:
> Since the system/default nodepool is already separate from the additional node_pool what is the reasoning behind allowing setting of the mode in the additional nodepools? They should always be User correct? This seems like it may be an unnecessary option.

@NormJohnIV I initially thought so too but then some usecases like performance validation, HA testing, deploy Viya4 monitoring into a separate "system" nodepool or when 'System' nodepool created by default_nodepool should fail, there needs to be a way for users create a new 'System' nodepool to cordon off all system pods without taking down Viya4 deployments and/or entire AKS cluster. Without providing the 'mode' I don't see another way, agree?

modules/azure_aks/main.tf
manoatsas: oops.. missed that the previous link was removed entirely, fixed it!

thpang:
After reading over how `system` node pools work with Azure and based on information from this article : 

https://pumpingco.de/blog/modify-aks-default-node-pool-in-terraform-without-redeploying-the-cluster/

It's best we do not expose the `mode` flag at this time. This is something the kubernetes admin would be taking care of and would possibly do outside of terraform and the sync later. This addition would need more investigation.

@manoatsas if you can make that last change we can merge this pr. Thx.

thpang (commented):
Changes look good.

thpang (approved):

PR #88, opened 2021-01-21
Updated provider and resource versions.

thpang:


thpang:
I have updated the version per the docs. @NormJohnIV if you can review.

PR #92, opened 2021-01-25
Removed auto ssh key gen if not supplied. Variable is now required.

thpang:


docs/CONFIG-VARS.md
```diff
@@ -7,7 +7,6 @@ Supported configuration variables are listed in the table below.  All variables
 - [List of valid configuration variables](#list-of-valid-configuration-variables)
   - [Table of Contents](#table-of-contents)
   - [Required Variables](#required-variables)
-    - [Application](#application)
```
manoatsas, line 10: Is there a reason as to why the link to "Application" was removed?

docs/CONFIG-VARS.md
```diff
@@ -28,13 +27,11 @@ Terraform input variables can be set in the following ways:
 
 ## Required Variables
 
-### Application
```
manoatsas, line 31: Ditto

docs/CONFIG-VARS.md
```diff
@@ -28,13 +27,11 @@ Terraform input variables can be set in the following ways:
...
-
 | Name | Description | Type | Default | Notes |
 | :--- | ---: | ---: | ---: | ---: |
 | prefix | A prefix used in the name of all the Azure resources created by this script. | string | | The prefix string must start with a lowercase letter and contain only alphanumeric characters and dashes (-), but cannot end with a dash. |
 | location | The Azure Region to provision all resources in this script | string | "East US" | |
-| tags | Map of common tags to be placed on all Azure resources created by this script | map | { project_name = "sasviya4", environment = "dev" } | |
```
manoatsas, line 37: Ditto again, "tags" I believe are required for some Azure resources

manoatsas (changes requested):

docs/CONFIG-VARS.md
manoatsas: Looks like "tags" can be left out of required section.

docs/CONFIG-VARS.md
thpang: Yes, to keep it consistent with the AWS repo. If it's need we need to make changes across all providers.

docs/CONFIG-VARS.md
thpang: Same as above.

docs/CONFIG-VARS.md
thpang: So "tags" have are now in the "General" section.

PR #98, opened 2021-02-04
Node Pool label removal

thpang:


thpang:
Closing pulled code from older branch snapshot.

Issue #84, opened 2021-01-13
Expose NFS Disk Type Parameter


andybouts:
Please expose the NFS Disk Type parameter so a user can select a premium or ultra disk for a "standard" VM based NFS configuration.

PR #110, opened 2021-02-26
Vm disk config

### LINKED ISSUE #81: Expose NFS/Jumpbox machine_type
In order to change the instance type for the nfs and jumpbox, variables need to be exposed.

manoatsas:
closes #81 and #84
- surface variables to configure VM types for Jump & NFS
- changed defaults SKUs
- adjust other values for networking/cahching when using Ultra disk

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -183,6 +183,7 @@ module "jump" {
   azure_rg_name     = azurerm_resource_group.azure_rg.name
   azure_rg_location = var.location
   vnet_subnet_id    = data.azurerm_subnet.misc-subnet.id
+  machine_type      = var.jump_vm_machine_type
   azure_nsg_id      = azurerm_network_security_group.nsg.id
   tags              = var.tags
   create_vm         = local.create_jump_vm
@@ -190,6 +191,8 @@ module "jump" {
   ssh_public_key    = file(var.ssh_public_key)
   cloud_init        = var.storage_type == "dev" ? null : data.template_cloudinit_config.jump.rendered
   create_public_ip  = var.create_jump_public_ip
+
+  depends_on = [module.vnet]
 }
 
 resource "azurerm_network_security_rule" "jump-ssh" {
@@ -206,6 +209,8 @@ resource "azurerm_network_security_rule" "jump-ssh" {
   destination_address_prefix  = "*"
   resource_group_name         = azurerm_resource_group.azure_rg.name
   network_security_group_name = azurerm_network_security_group.nsg.name
+
+    depends_on = [module.jump]
 }
 
 data "template_file" "nfs-cloudconfig" {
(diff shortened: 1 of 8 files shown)


main.tf
```diff
@@ -188,7 +189,7 @@ module "jump" {
   cloud_init        = data.template_cloudinit_config.jump.rendered
   create_public_ip  = var.create_jump_public_ip
 
-  depends_on = [module.nfs]
+  depends_on = [module.vnet]
```
thpang, line 192: Can you provide info on changing this one from nfs -> vnet? The nfs server must be active before the jump box can configure.

main.tf
```diff
@@ -242,6 +247,8 @@ resource "azurerm_network_security_rule" "vm-ssh" {
   destination_address_prefix  = "*"
   resource_group_name         = azurerm_resource_group.azure_rg.name
   network_security_group_name = azurerm_network_security_group.nsg.name
+
+  depends_on = [module.nfs]
```
thpang, line 251: What is the reason for adding this dependency on the nfs module? This rule is for all vm's created. Not sure why this is needed.

modules/azurerm_vm/variables.tf
```diff
@@ -65,16 +70,31 @@ variable data_disk_caching {
...
+  description = "The Type of Storage Account which should back this the Internal OS Disk. Possible values are Standard_LRS, StandardSSD_LRS and Premium_LRS. Changing this forces a new resource to be created"
+}
+
 variable os_disk_caching {
   default = "ReadOnly"
 }
 
 variable enable_accelerated_networking {
-  default = true
+  default = false
```
thpang, line 97: Why would this default to false here? The thought would be the customer would want this by default and if they did not need it they would turn it off.

main.tf
manoatsas: `vnet` to add explicit module dependency and issues encountered when testing. If jump vm depends on 'nfs` then I can add it back with a comment.

main.tf
manoatsas: To add explicit dependencies. There were collision with priority values, likely from prior to merge that merged nfs & jump nsg rules. I can test and remove.

modules/azurerm_vm/variables.tf
manoatsas: For jump vm types it had to be set false and for nfs it's handled here - https://github.com/sassoftware/viya4-iac-azure/blob/vm-disk-config/modules/azurerm_vm/main.tf#L19

main.tf
thpang: In the future you can add both since `depends_on` does take a list. Also the fact the Jump server does require the NFS server to be active was the original reason for the nfs dependency.

For example:
`depends_on = [module.nfs, module.vnet]`

PR #118, opened 2021-03-08
Fixes: #116 - nfs vm machine type has been corrected.

thpang:
**NOTE** - VM for NFS server was reverted back to its correct value in the docs.

### DIFF
--- a/docs/CONFIG-VARS.md
+++ b/docs/CONFIG-VARS.md
@@ -78,7 +78,7 @@ Ubuntu 18.04 LTS is the operating system used on the Jump/NFS servers. Ubuntu cr
 | create_jump_vm | Create bastion host | bool | true | |
 | create_jump_public_ip | Add public ip to jump VM | bool | true | |
 | jump_vm_admin | OS Admin User for the Jump VM | string | "jumpuser" | |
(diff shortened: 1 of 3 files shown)


docs/CONFIG-VARS.md
```diff
@@ -198,7 +198,7 @@ Note: The 128 default is in GB, so with a RAID5, the default is 4 disks, [so the
 | :--- | ---: | ---: | ---: | ---: |
 | create_nfs_public_ip | Add public ip to the NFS server VM | bool | false | |
 | nfs_vm_admin | OS Admin User for the NFS server VM | string | "nfsuser" | |
-| nfs_vm_machine_type | SKU to use for NFS server VM | string | "Standard_B2s_v3" | To check for valid types for your subscription `az vm list-skus --resource-type virtualMachines --subscription $subscription --location $location -o table`|
+| nfs_vm_machine_type | SKU to use for NFS server VM | string | "Standard_D8s_v4" | To check for valid types for your subscription `az vm list-skus --resource-type virtualMachines --subscription $subscription --location $location -o table`|
```
manoatsas, line 201: NFS VM type should be D-series (https://github.com/sassoftware/viya4-iac-azure/blob/main/variables.tf#L283). B-series is only for Jump VM.

thpang: Yes, that's is the default for the NFS VM. It was documented incorrectly. The adjustment here has fixed that issue.

manoatsas (approved):
LGTM

Issue #109, opened 2021-02-24
Add tags to Azure Container Registry resource


st-jan:
Please add the tags argument to the ACR resource block so that tags defined in the variables file get added to the resource in Azure. Currently, re-running Terraform marks the ACR as requiring an update, because automatically added tags are targeted for removal.

manoatsas:
Fixed in PR #123

Issue #108, opened 2021-02-22
NFS NIC error prevents Vnet/Subnet deletion on AKS updates


manoatsas:
On AKS updates, NFS NIC resource does not allow Vnet update

```
module.aks.azurerm_kubernetes_cluster.aks: Still destroying... [id=/subscriptions/XXXX--...nerService/managedClusters/mnm-tst-aks, 4m0s elapsed]
module.aks.azurerm_kubernetes_cluster.aks: Destruction complete after 4m3s
module.vnet.azurerm_virtual_network.vnet: Destroying... [id=/subscriptions/XXXX/resourceGroups/mnm-tst-rg/providers/Microsoft.Network/virtualNetworks/mnm-tst-vnet]

Error: Error deleting Virtual Network "mnm-tst-vnet" (Resource Group "mnm-tst-rg"): network.VirtualNetworksClient#Delete: Failure sending request: StatusCode=400 -- Original Error: Code="InUseSubnetCannotBeDeleted" Message="Subnet mnm-tst-misc-subnet is in use by /subscriptions/XXXX/resourceGroups/mnm-tst-rg/providers/Microsoft.Network/networkInterfaces/mnm-tst-nfs-nic/ipConfigurations/mnm-tst-nfs-ip_config and cannot be deleted. In order to delete the subnet, delete all the resources within the subnet. See aka.ms/deletesubnet." Details=[]
```

manoatsas:
Tried to add  `depends_on = [module.vnet]` to module.nfs that still does not seem to enforce decencies between NFS VM/NIC and VNet/Subnets.

This seems to be due to an update to tags that forces VNet Subnets. I found a similar issue open in the module repo - https://github.com/Azure/terraform-azurerm-vnet/issues/41 

```
  # module.vnet.azurerm_virtual_network.vnet must be replaced
-/+ resource "azurerm_virtual_network" "vnet" {
        address_space         = [
            "192.168.0.0/16",
        ]
        dns_servers           = []
      ~ guid                  = "e40fe24a-0818-4a80-94c9-0dce5f737f96" -> (known after apply)
      ~ id                    = "/subscriptions/XXXXX/resourceGroups/mnm-tst-rg/providers/Microsoft.Network/virtualNetworks/mnm-tst-vnet" -> (known after apply)
      ~ location              = "eastus2" -> (known after apply) # forces replacement
        name                  = "mnm-tst-vnet"
        resource_group_name   = "mnm-tst-rg"
[...]

Issue #106, opened 2021-02-19
Add dependency for kubernetes_config_map on AKS 


manoatsas:
When an error occurs on terraform (re)apply and destroy get this error - 
`Error: Get "http://localhost/api/v1/namespaces/kube-system/configmaps/sas-iac-buildinfo": dial tcp [::1]:80: connect: connection refused`

Any kubernetes resources should have dependency on K8s cluster resource and cleaned up accordingly.

thpang:
I'll look into this. Have not seen this error in our current testing. Thx.

manoatsas:
Yea it probably does not occur in a happy-path flow but when a re-apply fails, it where I hit this.
It seems to leave the resource in limbo and any further terraform commands are stopped in the tracks

```
terraform state list                                                      
data.azurerm_public_ip.aks_public_ip
data.azurerm_subnet.aks-subnet
data.azurerm_subnet.misc-subnet
data.azurerm_subscription.current
data.external.git_hash
data.external.iac_tooling_version
data.template_cloudinit_config.jump
data.template_cloudinit_config.nfs
data.template_file.jump-cloudconfig
data.template_file.nfs-cloudconfig
azurerm_network_security_group.nsg
azurerm_resource_group.azure_rg
kubernetes_config_map.sas_iac_buildinfo
module.jump.azurerm_public_ip.vm_ip[0]
module.nfs.azurerm_linux_virtual_machine.vm[0]
module.nfs.azurerm_managed_disk.vm_data_disk[0]
module.nfs.azurerm_managed_disk.vm_data_disk[1]
module.nfs.azurerm_managed_disk.vm_data_disk[2]
module.nfs.azurerm_managed_disk.vm_data_disk[3]
module.nfs.azurerm_network_interface.vm_nic[0]
module.nfs.azurerm_network_interface_security_group_association.vm_nic_sg[0]
module.nfs.azurerm_virtual_machine_data_disk_attachment.vm_data_disk_attach[0]
module.nfs.azurerm_virtual_machine_data_disk_attachment.vm_data_disk_attach[1]
module.nfs.azurerm_virtual_machine_data_disk_attachment.vm_data_disk_attach[2]
module.nfs.azurerm_virtual_machine_data_disk_attachment.vm_data_disk_attach[3]
module.vnet.data.azurerm_resource_group.vnet
module.vnet.azurerm_subnet.subnet[0]
module.vnet.azurerm_subnet.subnet[1]
[...]

PR #123, opened 2021-03-12
Fixed dependency issues

manoatsas:
close - #106 - remove kubernetes_provider that causes the issue and revert back to template file and null_resource
close - #108 - AKS updates no longer are triggered on changes to `tags`, more on that below, however in a few NFS update paths can hit - https://github.com/Azure/terraform-azurerm-compute/issues/107 
close - #109 - add tags to ACR

Added local module for azurerm_vnet to avoid https://github.com/Azure/terraform-azurerm-vnet/issues/41. In digging through this, it's an issue with multiple data source objects which doesn't have 'tags' attribute. TF sees this as a difference and with `depends_on` does a forced replacement of resources - Subnets, AKS etc. These changes solves any unwanted forced replacement and does an in place update to resources triggered due to changes to 'tags'


manoatsas:
Restored kubernetes provider and ran tests updating tags and disk_types that passes and does not hit the error with AKS updated in place. For any cases that destroy AKS outside of Terraform shall not be supported.

Issue #114, opened 2021-03-03
Creation with internal IP-addresses


jan-ehlers:
A customer requires us to use internal network addresses for all ressources in their Azure tenant. They use their Azure tenant as part of their internal network. I cannot find any variables to control the ip-address ranges that will be used for created ressources such as:
- Azure PostgreSQL
- AKS including nodes (AKS private cluster option?)

NormJohnIV:
postgres will be private if both postgres_public_access_cidrs and default_public_access_cidrs are empty. AKS cluster will be private if both default_public_access_cidrs and cluster_endpoint_public_access_cidrs are left empty. As for the viya/k8s ingress that is not part of the IAC code. It falls into the viya4-deployment and custom overlay realm

jan-ehlers:
Please also consider controlling the vnet for Azure Netapp Files with variables. We would like to link mount that to a jumphost in the customer's on-premises datacenter, or in another ressource group due to customers policies about no jumphosts in ressource groups with AKS..

PR #128, opened 2021-03-17
Support bring your own

### LINKED ISSUE #127: Deploy in existing (private) networks
It would be nice if there were an option to deploy the infrastructure into existing networks.  i.e., allow the user to supply vnet and subnet ids as input variables.  

This likely also means that an existing resource group would have to be used as well:  https://github.com/sassoftware/viya4-iac-azure/issues/68

Lastly, this should also allow for totally private networks - no public IPs anywhere.  https://github.com/sassoftware/viya4-iac-azure/issues/114 & https://docs.microsoft.com/en-us/azure/aks/ingress-internal-ip

NormJohnIV:
Add support for providing pre-existing rg, vnet and subnets
closes #127 #114

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -524,7 +524,7 @@ variable "subnets" {
   default = [
     {
       "name": "aks-subnet",
-      "prefixes": ["192.168.16.0/20"],
+      "prefixes": ["192.168.0.0/23"],
       "service_endpoints": ["Microsoft.Sql"],
       "enforce_private_link_endpoint_network_policies": false,
       "enforce_private_link_service_network_policies": false,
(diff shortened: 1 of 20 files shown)


variables.tf
```diff
@@ -497,3 +497,81 @@ variable "log_analytics_solution_promotion_code" {
...
+
+variable "vnet_name" {
+  type    = string
+  default = null
+  description = "Name of pre-exising vnet. Leave blank to have one created"
+}
+
+variable "vnet_address_space" {
+  type        = string
+  default     = "192.168.0.0/16"
```
thpang, line 516: Just to clarify this value replaces the old `10.0.0.0/16` CIDR range from the azure_vnet variables file.

NormJohnIV: yes

PR #148, opened 2021-05-12
Allow helper script to be sourced

andybouts:
Allow helper script to be sourced


andybouts:
Updated to Quote the $vrbl to resolve linter error

andybouts:
added error handling

PR #153, opened 2021-05-18
Upgrading to tf 0.15.3

thpang:
Updating tf and other versions to keep things up-to-date.


hahewlet:
I don't see a change in the readme text to indicate the new required version of terraform is 15.4.

thpang:
Docs are not complete at this time. Looking for functional testing. Terraform 0.15.4 is the only change that the end user will notice.

PR #161, opened 2021-06-09
Added in new minimal cluster option

thpang:


thpang:
@NormJohnIV if you can review this one it's ready. Thx.

variables.tf
```diff
@@ -581,3 +581,9 @@ variable "create_static_kubeconfig" {
   type        = bool
   default     = true
 }
+
+variable "cluster_node_pool_mode" {
+  description = "Flag for predefined cluster node configurations - Values : default, minimal"
+  type        = string
+  default     = "default"
```
RichWellum, line 588: So to create 'minimal' you would do: default = "minimal"? Seems like default is always a default and you need a var that says 'deployment type' - which defaults to default and can be set to 'minimum'?

RichWellum (changes requested):

PR #169, opened 2021-06-15
Add --output tsv to TF VAR tenant and subscription IDs

dixoncrews:
When following the instructions as written, the two variables had quotes around the values, which resulted in errors in the next steps. Adding the sed command to remove leading/trailing double quotes resolved the error.

### DIFF
--- a/docs/user/AzureHelpTopics.md
+++ b/docs/user/AzureHelpTopics.md
@@ -35,11 +35,11 @@ You can use the following commands, or alter them as needed to establish the env
 # for example, az account show --query '[environmentName, name, tenantId, user.name]'
 
 # set the tenant ID from a query; validate
-TF_VAR_tenant_id=$(sed -e 's/^"//' -e 's/"$//' <<< $(az account show --query 'tenantId'))
+TF_VAR_tenant_id=$(az account show --query 'tenantId' --output tsv)
 echo $TF_VAR_tenant_id
 
 # set the subscription ID from a query; validate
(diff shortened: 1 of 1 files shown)


thpang:
@dixoncrews won't be using sed. You'll have to find another way. sed changes based on Linux/Mac OS. Need something more generic.

thpang:
@andybouts this is one for you to review and verify.

enderm:
suggest adding "--output tsv", e.g.  
az account show --query tenantId --output tsv

dixoncrews:
@enderm Yep, that does it!

andybouts:
Hey @thpang , I can review the tf helper script, but I believe my tf helper script already does what they are doing.  I certainly did not use sed,  but instead parsed the output for a value.

andybouts:
Hey @dixoncrews , long time since we last spoke.  We might want to review and make sure the tf helper script I provided does the same thing.  I didn’t like keeping track of the tf env vars so I wrote the script to make it easier.

thpang:
@andybouts you are correct the code has that option already set. Thx.

andybouts:
All, [suggest you use this script ](https://github.com/sassoftware/viya4-iac-azure/blob/main/files/tools/terraform_env_variable_helper.sh) ... it works well and can be run as often as you want ... I also have suggested commands at the bottom that print out in case your credential needs to be reset, which happens enough to me that it's annoying.  This script already included the changes that you pushed to the md.

PR #171, opened 2021-06-21
Flex psql

### LINKED ISSUE #79: Support Azure Database for PostgreSQL - Flexible Server
Support https://docs.microsoft.com/en-us/azure/postgresql/flexible-server/overview which supports stop/start features for cost saving measures.
Implementation depends on -https://github.com/terraform-providers/terraform-provider-azurerm/issues/8654

manoatsas:
close #79

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -38,7 +38,7 @@ locals {
   kubeconfig_filename = "${var.prefix}-aks-kubeconfig.conf"
   kubeconfig_path     = var.iac_tooling == "docker" ? "/workspace/${local.kubeconfig_filename}" : local.kubeconfig_filename
 
-  subnets = { for k, v in var.subnets : k => v if ! ( k == "netapp" && var.storage_type == "standard")}
+  subnets = { for k, v in var.subnets : k => v if ! (k == "netapp" && var.storage_type == "standard") }
 
   container_registry_sku = title(var.container_registry_sku)
 }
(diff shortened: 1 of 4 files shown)


main.tf
```diff
@@ -145,7 +145,7 @@ data "template_cloudinit_config" "nfs" {
 }
 
 module "nfs" {
-  source                         = "./modules/azurerm_vm"
+  source = "./modules/azurerm_vm"
```
thpang, line 148: Don't lint the files. There are several issues/prs where you're run `tf format` or something similar and it creates a lot of changes. Formatting should be done by hand. This creates lots of noise in the issue/pr with changes.

thpang (commented):
Added a comment. Should refrain from `tf fmt` to much noise in changes. Also will this model work across the other providers? Thinking the variables should be `create_postgres` and then possibly `postgres_type` where we identify what type of postgres server is being created. This could help simplify the conditional logic.

main.tf
manoatsas: I'd suggest otherwise - i.e., require 'terraform fmt` prior to all commits. It uses Terraform formatting Vs. custom/editor formatting that helps with readability of the code base.

manoatsas:
> Added a comment. Should refrain from `tf fmt` to much noise in changes. Also will this model work across the other providers? Thinking the variables should be `create_postgres` and then possibly `postgres_type` where we identify what type of postgres server is being created. This could help simplify the conditional logic.

This Flexible Server is Azure specific offering that other providers don't seem to have yet, so not sure who this cuts across other IaC projects. As for switching to `postgres_type` I'd agree it'll be up to the user to enter a valid string to match with the type of Postgres.

thpang:
Will pull from draft once it's out of preview.

manoatsas:
Closing in lieu of https://github.com/sassoftware/viya4-iac-azure/pull/220

PR #175, opened 2021-06-29
Support spot instances

manoatsas:


riragh:
At this time SAS does not support spot VM's. In the future if this is a requirement we'll revisit.

PR #183, opened 2021-07-27
resolve AZ CLI Change

andybouts:
AZ CLI has a breaking change ... the updates to the script worked as expected for myself and Tom S

### DIFF
--- a/files/tools/terraform_env_variable_helper.sh
+++ b/files/tools/terraform_env_variable_helper.sh
@@ -35,13 +35,13 @@ echo -e "\nTF_VAR_subscription_id:"
 echo -e "$TF_VAR_subscription_id"
 
 # obtain the client ID
-TF_VAR_client_id=$(az ad sp show --id http://"$YOURSP" --query appId --output tsv)
+TF_VAR_client_id=$(az ad sp list --display-name "$YOURSP" --query '[].appId' --output tsv)
 export TF_VAR_client_id
 echo -e "\nTF_VAR_client_id for this Service Principal $YOURSP:"
 echo -e "$TF_VAR_client_id"
 
 # obtain a client secret
-TF_VAR_client_secret=$(az ad sp create-for-rbac --skip-assignment --name http://"$YOURSP" --query password --output tsv)
+TF_VAR_client_secret=$(az ad sp create-for-rbac --skip-assignment --name "$YOURSP" --query password --output tsv)
 export TF_VAR_client_secret
 echo -e "\nTF_VAR_client_secret:"
 echo -e "$TF_VAR_client_secret"


thpang:
Would you please explain what is being fixed here? I've run the old and new commands and found they both return the same value.

andybouts:
The primary breaking change is at line 44.  Running it the previous way with this parameter `--name http://"$YOURSP"`would create a SP named `$YOURSP`.  As of today though, running that previous command creates a different service principal named `"http://$YOURSP"`:

```
anbxx@cldxxx:~/blue-xxx_oddMonths/viya4-iac-azure$ az ad sp list --show-mine --query '[].displayName' | grep SP
  "{truncated}_Advisory_SP",
  "http://{truncated}_Advisory_SP"
```

While I was working through resolving the breaking change with the name prepended with `http://`, I learned that the `az ad sp list` command has more options and is more robust than the `az ad sp show` command.  That said, technically speaking, I believe the old and the version of row 38 may produce the same results ...

thpang:
What version of the az cli is associated with this change? Given this code has gone out making the change may break if folks are running the older version of the az cli.

andybouts:
I developed the test and validated it on the following.  I don't know the previous version of the AZ CLI that the script was originally developed on.  I know that when I attempted to build the IAC yesterday for the first time in about 5-6 weeks, I was not able to build one because the same commands that worked before created a new SP instead of reusing my existing one.

We could ask IT when it was last updated on this shared jump host:

```anbxxx@cldlgxxx:~$ az --version
azure-cli                         2.26.1

core                              2.26.1
telemetry                          1.0.6

Python location '/opt/az/bin/python3'
```

files/tools/terraform_env_variable_helper.sh
```diff
@@ -35,13 +35,13 @@ echo -e "\nTF_VAR_subscription_id:"
...
 # obtain the client ID
-TF_VAR_client_id=$(az ad sp show --id http://"$YOURSP" --query appId --output tsv)
+TF_VAR_client_id=$(az ad sp list --display-name "$YOURSP" --query '[].appId' --output tsv)
 export TF_VAR_client_id
 echo -e "\nTF_VAR_client_id for this Service Principal $YOURSP:"
 echo -e "$TF_VAR_client_id"
 
 # obtain a client secret
-TF_VAR_client_secret=$(az ad sp create-for-rbac --skip-assignment --name http://"$YOURSP" --query password --output tsv)
+TF_VAR_client_secret=$(az ad sp create-for-rbac --skip-assignment --name "$YOURSP" --query password --output tsv)
```
thpang, line 44: Do see where the old command does now set the name to include `http://` with your Service Principal name. Would be nice to know what/how this changed. If you can find docs and or info that would be helpful. We'll need to adjust other items in the `How-to` doc for this project as well when creating a service principal.

thpang (commented):
Added comments for review.

files/tools/terraform_env_variable_helper.sh
```diff
@@ -35,13 +35,13 @@ echo -e "\nTF_VAR_subscription_id:"
 echo -e "$TF_VAR_subscription_id"
 
 # obtain the client ID
-TF_VAR_client_id=$(az ad sp show --id http://"$YOURSP" --query appId --output tsv)
+TF_VAR_client_id=$(az ad sp list --display-name "$YOURSP" --query '[].appId' --output tsv)
```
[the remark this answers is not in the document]
andybouts: Yes, [az ad sp list](https://docs.microsoft.com/en-us/cli/azure/ad/sp?view=azure-cli-latest#az_ad_sp_list) is more robust with parameters like `--filter`, `--show-mine`, etc, so I believe it's a better command to call (although in this instance the commands are essentially the same), having list allows the user more flexibility if they wanted to run these commands on an ad-hoc basis.

files/tools/terraform_env_variable_helper.sh
andybouts: According to the [documentation for the az ad sp module](https://docs.microsoft.com/en-us/cli/azure/ad/sp?view=azure-cli-latest#az_ad_sp_create_for_rbac), it seems to always use the latest version instead of specific versions, so after some searching, I was not able to pinpoint this to a specific change in the AZ CLI from version to version.  Perhaps, it would be possible to search through the[ GitHub project ](https://github.com/Azure/azure-cli) and find this, but I did not have time for that.

I just know that this new command works as expected with the version of the AZ CLI that I mentioned above for me as well as a colleague.  

If you would need, I could add a usage note or another echo in the script with the version number this was developed and validated using.

thpang: No need. Just need to adjust our auth docs now.

thpang (approved):
Looks good and thanks for the follow-up.

PR #191, opened 2021-08-13
Support BYO vnet located outside the AKS RG

bernardmaltais:
This pull request is to allow users of this repo that have an existing vnet in a resource group that is different from the target resource group of the Viya4 AKS cluster and other associated resources.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -58,7 +58,7 @@ module "vnet" {
 
   name                = var.vnet_name
   prefix              = var.prefix
-  resource_group_name = module.resource_group.name
+  resource_group_name = var.vnet_resource_group_name == null ? module.resource_group.name : var.vnet_resource_group_name
   location            = var.location
   subnets             = local.subnets
   existing_subnets    = var.subnet_names
--- a/variables.tf
+++ b/variables.tf
@@ -511,6 +511,12 @@ variable "resource_group_name" {
   description = "Name of pre-exising resource group. Leave blank to have one created"
 }
 
+variable "vnet_resource_group_name" {
+  type    = string
+  default = null
+  description = "Name of pre-exising vnet resource group. Leave blank to have one created"
+}
+
 variable "vnet_name" {
   type    = string
   default = null


main.tf
```diff
@@ -58,7 +58,7 @@ module "vnet" {
 
   name                = var.vnet_name
   prefix              = var.prefix
-  resource_group_name = module.resource_group.name
+  resource_group_name = var.vnet_resource_group_name == null ? module.resource_group.name : var.vnet_resource_group_name
```
thpang, line 61: Though you call it `vnet_resource_group_name` it could be any resource group. So the naming should change.

variables.tf
```diff
@@ -511,6 +511,12 @@ variable "resource_group_name" {
   description = "Name of pre-exising resource group. Leave blank to have one created"
 }
 
+variable "vnet_resource_group_name" {
```
thpang, line 514: Same comment here could be any resource group so name should change.

thpang (commented):
If you can review my comments. Thx.

bernardmaltais:
Should I update the documentation to explain how to use this feature such that users can know about it and properly use it?

docs/CONFIG-VARS.md
```diff
@@ -122,6 +122,7 @@ the existing resources:
 | Name | Description | Type | Default | Notes |
 | :--- | ---: | ---: | ---: | ---: |
 | resource_group_name | Name of pre-existing resource group | string | null | Only required if deploying into existing resource group. |
+| vnet_resource_group_name | Name of a pre-exising resource group containing the BYO vnet resource | string | null | Only required if deploying into existing vnet that is not located in the resource_group_name specified above |
```
thpang, line 125: @bernardmaltais given the extra resource group, is there a need for vnet/rg peering information here? Wanting to make sure there are not connectivity issues if one uses the current rg and the added `vnet_resource_group` here.

variables.tf
```diff
@@ -511,6 +511,12 @@ variable "resource_group_name" {
...
+  type    = string
+  default = null
+  description = "Name of a pre-exising resource group containing the BYO vnet resource. Leave blank if you are not using a BYO vnet or if the BYO vnet is co-located with the SAS Viya4 AKS cluster."
```
thpang, line 517: I would change the wording here. Remove the Leave blank wording. If it's not set it's ignored. Nothing more to say here.

thpang (changes requested):
Just a couple of adjustments on documentation

thpang (approved):

PR #192, opened 2021-08-15
Add support for pre-existing NFS server IP

bernardmaltais:
This pull request allow users to specify the pre-existing NFS server IP address to use as part of the deployment.


riragh:
At this time adding support for BYO NFS server is not priority but will be investigated in future.

PR #196, opened 2021-09-27
fix(tf): allow override of egress and fmt

onyxhat:
- allow explicit override of cluster egress
- handle certs inline/as secrets
- update output if no public ip is present


thpang:
@onyxhat thx for the PR. Reviewing. Most of the changes I am see so far are from you running `terraform fmt` or some other formatting tool on the code base. Will have to digest what's actually changed vs just formating adjustments.

onyxhat:
I can go back and remove the formatting - I didn't intend to do that, but I'm running a vscode extension that automatically does that on save. Apologies on that.

thpang:
No worries. Either way I'll be able to get through the code.

thpang:
@GonzaloUlla you will need to backout the formatting changes. These really change the way not only change the code base and it's layout but it also changes the examples files, etc. At some some point in the future we need to adopt using a formatting tool; however, not today. Once you've made those changes I'll scan through the code.

onyxhat:
can do - I'll get those changes in (unformatted) today

onyxhat:
This has been replaced with #197

PR #197, opened 2021-10-01
fix(tf): allow overrides

onyxhat:
- allow aks to use service principle
- allow overriding vnet_rgrp
- allow overriding egress
- handle certificates inline/as secrets
- handle no output on public_ip

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -63,7 +63,7 @@ module "vnet" {
 
   name                = var.vnet_name
   prefix              = var.prefix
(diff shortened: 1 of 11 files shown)


onyxhat:
Fixed

onyxhat:
@thpang requested changes have been made

thpang:
@manoatsas and @enderm if you could both review the code changes here. Had a few initial passes. It's ready now for a larger audience. Thx.

PR #198, opened 2021-10-05
BYO AKS identity

enderm:
- add ability to Bring Your Own AKS identity (independent of private_cluster)
- integrate BYO resource group (currently based on PR #191)
- documente "aks_uai_name" variable
- add new `aks_identity` variable do support using Service Principal as AKS Identity 
- refactor resource group and nsg modules into the locals/resource/data pattern made possible by tf 13 lazy conditional evaluation


iam.tf
```diff
@@ -0,0 +1,26 @@
+data "azurerm_user_assigned_identity" "uai" {
+  count               = var.aks_identity == "uai" ? ( var.aks_uai_name == null ? 0 : 1 ) : 0
+  name                = var.aks_uai_name
+  resource_group_name = local.network_rg.name
+}
+
+resource "azurerm_user_assigned_identity" "uai" {
+  count               = var.aks_identity == "uai" ? ( var.aks_uai_name == null ? 1 : 0 ) : 0
```
thpang, line 8: Should the last 0 (zero) actually be a 1 with the this logic? I am reading,

if aks_identity == uai then
  if aks_uai_name is null then
    count = 1
  else
    count = 0
else 
  count = 0 <-- I believe this should be 1 given the identity is set so it should be created.

iam.tf
```diff
@@ -0,0 +1,26 @@
...
+}
+
+# need to be able to read and write custom route table in the BYO network
+resource "azurerm_role_assignment" "uai_byo_rg_role" {
+  count                = ( var.aks_identity == "uai"
+                           ? ( var.aks_uai_name == null
+                               ? ( var.vnet_name == null ? 0 : 1 )
+                               : 0
+                             )
+                           : 0
```
thpang, line 21: Is this correct as well? Again, just not sure what the value of aks_identity actually is setting and how it generates the creation of items.

main.tf
```diff
@@ -30,46 +30,56 @@ provider "kubernetes" {
...
-  depends_on          = [module.resource_group]
+  resource_group_name = local.aks_rg.name
+  tags                = var.tags
 }
 
-module "nsg" {
-  source              = "./modules/azurerm_network_security_group"
-  prefix              = var.prefix
-  name                = var.nsg_name
+resource "azurerm_network_security_group" "nsg" {
```
thpang, line 58: Was the removal of the module for logic? Not sure why it was there and now it's been removed.

main.tf
```diff
@@ -97,12 +107,12 @@ module "jump" {
...
   name              = "${var.prefix}-jump"
-  azure_rg_name     = module.resource_group.name
+  azure_rg_name     = local.aks_rg.name
   azure_rg_location = var.location
   vnet_subnet_id    = module.vnet.subnets["misc"].id
   machine_type      = var.jump_vm_machine_type
-  azure_nsg_id      = module.nsg.id
-  tags              = module.resource_group.tags
+  azure_nsg_id      = local.nsg.id
+  tags              = var.tags
```
thpang, line 115: So between this item and line 150 below we use `vars` for some items and `locals` for others. Is there a reason it's not more consistent? Either one or the other. Seems like the logic flow is a bit scattered here.

main.tf
```diff
@@ -137,13 +147,13 @@ module "nfs" {
...
-  azure_rg_name                  = module.resource_group.name
+  azure_rg_name                  = local.aks_rg.name
   azure_rg_location              = var.location
   proximity_placement_group_id   = element(coalescelist(azurerm_proximity_placement_group.proximity.*.id, [""]), 0)
   vnet_subnet_id                 = module.vnet.subnets["misc"].id
   machine_type                   = var.nfs_vm_machine_type
-  azure_nsg_id                   = module.nsg.id
-  tags                           = module.resource_group.tags
+  azure_nsg_id                   = local.nsg.id
+  tags                           = var.tags
```
thpang, line 156: This seems as though it was set in the module and now it's a full blown var. Can you clarify?

thpang (changes requested):
Some general comments. Like the use of locals. The use of `var.tags` seems out of place or maybe it's there because of the removal of the modules. Why were the modules removed?

iam.tf
enderm: the new variable aks_identity=sp|uai tells us whether the user wants to use the Service Principal or an Identity. So, if aks_identity is NOT UAI, then we don't want create an UAI here.

iam.tf
enderm: again, if aks_identity is "sp", then we don't create uai's or role assignments

main.tf
enderm: removed the module for simplicity and consistency. With the additional vnet_rg it would have meant adding a couple more input/output vars to the module. That would have been more code than using the local/data/resource pattern.

main.tf
enderm: generally, if the vars can be used as-is, without further logic, we use it.
If we need to add conditionals (there are - unfortunately - some dependencies between input vars), or wrapping logic (like data vs. resource for the byo items), we tend to use locals.

main.tf
enderm: not sure what "it" refers to here.

iam.tf
thpang: Got it.

main.tf
thpang: It was the module.resource_group but you answered the vars questions above.

PR #204, opened 2021-10-26
add new byo options to examples

enderm:


thpang (commented):
Do we have docs and examples outlining this change? Asking as the CONFIG-VARS.md seems to have these items but the user/BringYourOwn.md does not seem to be updated.

enderm:
this PR is a followup to https://github.com/sassoftware/viya4-iac-azure/pull/198 . Somebody pointed out that we had forgotten to update the byo example file (it was likely you - I cannot find the exchange right now).
Revamping the byo/private cluster readme will come as part of the upcoming change of making private clusters official.

thpang:
Ok, pushing into `main` need to talk about creating a release for this one, then moving on to AWS/GCP items.

PR #207, opened 2021-10-29
make ssh_public_key optional 

enderm:


main.tf
```diff
@@ -139,7 +138,7 @@ module "aks" {
   aks_cluster_os_disk_size                 = var.default_nodepool_os_disk_size
   aks_cluster_node_vm_size                 = var.default_nodepool_vm_type
   aks_cluster_node_admin                   = var.node_vm_admin
-  aks_cluster_ssh_public_key               = file(var.ssh_public_key)
+  aks_cluster_ssh_public_key               = local.ssh_public_key
```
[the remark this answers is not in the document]
enderm: imo there are two ways we can go here:
- always check for a valid key file (i.e. remove the "var.create_jump_vm || var.storage_type == "standard"" on the local)
- silently set the key value to "null" if the keyfile is not valid (using try())

Preference?

examples/sample-input-byo.tfvars
```diff
@@ -6,7 +6,6 @@
 # These required variables' values MUST be provided by the User
 prefix         = "<prefix-value>"
 location       = "<azure-location-value>" # e.g., "eastus2"
-ssh_public_key = "~/.ssh/id_rsa.pub"
```
thpang, line 9: @enderm if you can add the `ssh_public_key` entries back into the sample files. As we chatted a bit on Friday, I believe the variable `ssh_public_key` needs to be present in order for people to see it and know what to change. Especially since it controls access to their VM's. Might need to move the variables location like you have in the doc.

examples/sample-input-defaults.tfvars
```diff
@@ -6,7 +6,6 @@
 # These required variables' values MUST be provided by the User
 prefix   = "<prefix-value>" # this is a prefix that you assign for the resources to be created
 location = "<azure-location-value>" # e.g., "eastus2"
-ssh_public_key = "<path-to-public-key-on-host-running-iac>" # Name of file with public ssh key for connecting to the VMs
```
thpang, line 9: Same as above. Also change the value to match the byo entry.

examples/sample-input-ha.tfvars
```diff
@@ -6,7 +6,6 @@
 # These required variables' values MUST be provided by the User
 prefix   = "<prefix-value>" # this is a prefix that you assign for the resources to be created
 location = "<azure-location-value>" # e.g., "eastus2"
-ssh_public_key = "<path-to-public-key-on-host-running-iac>" # Name of file with public ssh key for connecting to the VMs
```
thpang, line 9: Same as above. Also change the value to match the byo entry.

examples/sample-input-minimal.tfvars
```diff
@@ -6,7 +6,6 @@
 # These required variables' values MUST be provided by the User
 prefix   = "<prefix-value>" # this is a prefix that you assign for the resources to be created
 location = "<azure-location-value>" # e.g., "eastus2"
-ssh_public_key = "<path-to-public-key-on-host-running-iac>" # Name of file with public ssh key for connecting to the VMs
```
thpang, line 9: Same as above. Also change the value to match the byo entry.

examples/sample-input-ppg.tfvars
```diff
@@ -6,7 +6,6 @@
 # Values for these required variables MUST be provided
 prefix   = "<prefix-value>" # this is a prefix that you assign for the resources to be created
 location = "<azure-location-value>" # e.g., "eastus2"
-ssh_public_key = "<path-to-public-key-on-host-running-iac>" # Name of file with public ssh key for connecting to the VMs
```
thpang, line 9: Same as above. Also change the value to match the byo entry.

thpang (changes requested):
Updates to the example files.

locals.tf
```diff
@@ -6,6 +6,11 @@ locals {
   create_jump_public_ip = var.create_jump_public_ip == null ? local.is_standard : var.create_jump_public_ip
   create_nfs_public_ip  = var.create_nfs_public_ip == null ? local.is_standard : var.create_nfs_public_ip
 
+  ssh_public_key = ( var.create_jump_vm || var.storage_type == "standard"
+                     ? file(var.ssh_public_key)
+                     : null
+                   )
+
```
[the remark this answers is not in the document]
enderm: yup, that's what this change is all about - avoid unnecessary checks.
Lazy evaluations will prevent file() to be executed if the condition is not met. 
Otherwise, file() will give a pretty helpful error message.

main.tf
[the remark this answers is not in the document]
enderm: thoughts?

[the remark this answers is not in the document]
thpang: So you can't just pass in null into the module. The module has a section for `linux_profile` which setups node access. Going to ask the K8s-SIG on this one given it's related to the daemonset question. Short term for this PR, you'll need to alter code for the `linux_profile` this block should be a dynamic block based on the key value. Docs on that block are [here](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster#linux_profile).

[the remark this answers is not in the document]
enderm: that sounds like you are preferring the second option. Ok.

main.tf
```diff
@@ -139,7 +138,7 @@ module "aks" {
   aks_cluster_os_disk_size                 = var.default_nodepool_os_disk_size
   aks_cluster_node_vm_size                 = var.default_nodepool_vm_type
   aks_cluster_node_admin                   = var.node_vm_admin
-  aks_cluster_ssh_public_key               = file(var.ssh_public_key)
+  aks_cluster_ssh_public_key               = try( file(var.ssh_public_key), "")
```
thpang, line 141: Need to verify this will work if the file does not populate the value. This input variable is required in the module so having it set to "" or even null may be an issue.

thpang (commented):
Validation needed.

main.tf
enderm: I did change the module code to deal with an empty value. Pls have a look at the change in modules/azure_aks/main.tf  in this PR.

thpang (approved):
Looks good.

PR #210, opened 2021-11-08
Private cluster api endpoint

### LINKED ISSUE #158: Private cluster creation

enderm:
- add cluster_api_mode variable
 -  clarify *access_cidr usage for private/public endpoints

fixes #158


docs/CONFIG-VARS.md
```diff
@@ -54,22 +54,26 @@ For recommendations on how to set these variables in your environment, see [Auth
...
-By default, the API of the Azure resources that are being created is only accessible through authenticated Azure clients (such as the Azure Portal,
-the `az` CLI, the Azure Shell, etc.).
-To allow access for other administrative client applications (for example `kubectl`, `psql`, etc.), you must open the Azure firewall to allow access from your source
-IP addresses.
-To do this, specify ranges of IP addresses in [CIDR notation](https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing).
-Contact your Network Administrator to find the public CIDR range of your network.
+By default, the public endpoints of the Azure resources that are being created
+are only accessible through authenticated Azure clients
+(such as the Azure Portal, the `az` CLI, the Azure Shell, etc.).
+To allow access for other administrative client applications (for example `kubectl`, `psql`, `ssh` etc.), you can set Network Security Group (NSG) rules to control access from your source IP addresses.
```
thpang, line 60: So there is no longer an option for Azure Firewall as previously stated. This has been removed and you must only use a NSG?

docs/CONFIG-VARS.md
```diff
@@ -54,22 +54,26 @@ For recommendations on how to set these variables in your environment, see [Auth
...
+
+To do set these permissions as part of this Terraform script, specify ranges of IP addresses in [CIDR notation](https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing) with the following variables.
+
+NOTE: When deploying infrastructure into a private network (e.g. a VPN), with no public endpoints, the options documented in this block are not applicable.
+
+NOTE: The script will either create a new NSG, or use an existing NSG, if specified in the [`nsg_name`](#use-existing) variable.
```
thpang, line 66: Again, no Azure Firewall, this has been replaced here. Correct?

docs/CONFIG-VARS.md
```diff
@@ -54,22 +54,26 @@ For recommendations on how to set these variables in your environment, see [Auth
...
-| postgres_public_access_cidrs | IP address ranges allowed to access the Azure PostgreSQL Server | list of strings || Opens port 5432 by adding Ingress Rule on the NSG |
-| acr_public_access_cidrs | IP address ranges allowed to access the ACR instance | list of strings |||
+| cluster_endpoint_public_access_cidrs | IP address ranges allowed to access the AKS cluster API | list of strings | | For client admin access to the cluster api (by `kubectl`, for example). Only used with `cluster_api_mode=public`|
+| vm_public_access_cidrs | IP address ranges allowed to access the VMs | list of strings | | Opens port 22 for SSH access to the jump server and/or NFS VM by adding Ingress Rule on the NSG. Only used iwth `create_jump_public_ip=true` or `create_nfs_public_ip=true`   |
+| postgres_public_access_cidrs | IP address ranges allowed to access the Azure PostgreSQL Server | list of strings || Opens port 5432 by adding Ingress Rule on the NSG. Only used when creating postgres instances. |
+| acr_public_access_cidrs | IP address ranges allowed to access the ACR instance | list of strings || Only used with `create_container_registry=true` |
```
thpang, line 76: Are there guards on these variables later in the code to keep a user from setting these when not needed?

locals.tf
```diff
@@ -1,22 +1,17 @@
 locals {
   
   # Useful flags
-  is_private            = var.infra_mode == "private" ? true : false
-  is_standard           = var.infra_mode == "standard" ? true : false
-  create_jump_public_ip = var.create_jump_public_ip == null ? local.is_standard : var.create_jump_public_ip
-  create_nfs_public_ip  = var.create_nfs_public_ip == null ? local.is_standard : var.create_nfs_public_ip
```
thpang, line 7: How do I turn on/off the jump and nfs public IP values? Do I add these variables directly to my terraform vars file? No doc showing how this has changed.

thpang (changes requested):
Items to review

docs/CONFIG-VARS.md
enderm: The previous version did use the term "firewall", but it was used as a generic term, not as a reference to the "Azure Firewall" object. I want to argue that being more specific about how we do the access control, we avoid possible misunderstandings, and make it clearer how the _access_cidr variables are used.

docs/CONFIG-VARS.md
enderm: again, Azure Firewall was never in play here.

docs/CONFIG-VARS.md
enderm: cluster_api: https://github.com/sassoftware/viya4-iac-azure/blob/private-cluster/locals.tf#L11
vms: https://github.com/sassoftware/viya4-iac-azure/blob/private-cluster/vms.tf#L118
postgres: https://github.com/sassoftware/viya4-iac-azure/blob/private-cluster/locals.tf#L23
acr: https://github.com/sassoftware/viya4-iac-azure/blob/private-cluster/main.tf#L118

locals.tf
enderm: The intent is that you can set var.create_jump_public_ip and var.create_nfs_public_ip independently of the cluster endpoint.

thpang: This code has been released with the `infra_mode` flag already. So somewhere this needs to be addressed. This is going to be a breaking change for folks who may use this feature.  Need to have some docs on this one.

thpang (changes requested):
Need to have a location that outlines that `infra_mode` is going away and what's replacing it. Also how one will actually trigger the creation of the JUMP/NFS VM's as this is changing as well.

locals.tf
enderm: since infra_mode was never documented (and never integrated with viya4-deployment), maybe a note in one of our internal Teams channels would suffice here? In short, to get parity with infra_mode="private" you would now set:
cluster_api_mode="private"
create_jump_public_ip=false
create_nfs_public_ip=false (if storage_type="standard")

and possibly
default_public_access_cidrs=[]

thpang: I am ok with this. Need to set these create_<vm>_public_ip items to `true` for jump server by default in variable file and `false` for nfs server by default. Right now they are both `null` which was used when we were using the local value. All example files should be updated to explicitly set these as well. I believe the CONFIG-VAR.md doc is correct. I think the variables.tf file is not.

thpang (changes requested):
Small change in `variables.tf` file for jump/nfs server public ip values to match doc.

locals.tf
enderm: fixed variable.tf defaults. 
Not sure about the examples. We already set both variables explicitly in most of them. The behavior does not really change here.

variables.tf
```diff
@@ -225,7 +227,7 @@ variable "create_jump_vm" {
 }
 
 variable "create_jump_public_ip" {
-  default = null
+  default = false
```
thpang, line 230: This one is usually `true`, by default we should always stand up infra that can work, test with the [defaults](https://github.com/sassoftware/viya4-iac-azure/blob/main/examples/sample-input-defaults.tfvars) example file.

thpang (changes requested):
Small change and then verification needed.

variables.tf
enderm: fixed and tested.

thpang:
Ready to merge.

thpang (approved):
LGTM

PR #212, opened 2021-11-18
Update Dockerfile - missing bash 

tsej:
> No such file or directoryfiles/tools/iac_git_info.sh": env: can't execute 'bash

> No such file or directoryfiles/tools/iac_tooling_version.sh": env: can't execute 'bash


thpang:
Don't see how this helps? Running the container already has bash and the error is saying your container does not have the tooling. Not really sure because it looks like you copy and pasted some section of the error. Need more info to show this is valid.

tsej:
I'm sorry my mistake.

PR #214, opened 2021-11-22
(IAC-373) rename to cluster_egress_type

enderm:


variables.tf
```diff
@@ -513,7 +513,7 @@ variable "nsg_name" {
 variable "egress_public_ip_name" {
   type        = string
   default     = null
-  description = "Name of pre-existing Public IP for the Network egress."
+  description = "DEPRECATED: Name of pre-existing Public IP for the Network egress."
```
thpang, line 516: Just clarifying here. You've marked this as deprecated and left the variable in the code base so that current users of this variable will not see an error, correct? However, I don't see where this variable is actually used at this point. I would think if we had an old deprecated variable if set it would then be mapped into the new variable.

thpang (commented):
Looking for clarification on new variable. Thx.

variables.tf
enderm: yes, it is only left in so any code that uses it will not break.
I think what I am hearing you suggest is  that if cluster_egress_type is not set, but egress_public_ip_name IS set, then cluster_egress_type should internally be set to "userDefinedRouting". Correct? If so, I can make that change.

thpang: Yeah, I just think if we have it their deprecated, then the user should still be able to use the value. Also might want to let folks know that the next major release of the code the variable will disappear and not be valid. Thx.

riragh:
Testing complete. Tested two scenarios: 
1. Default where cluster_egress_type is set to 'loadbalancer' 
2. cluster_egress_type is set to "userDefinedRouting", with BYO setup.

sayeun (approved):
+2 👍

PR #217, opened 2022-01-25
(IAC-395) Updated default Kubernetes version from 1.19 -> 1.21

riragh:
### Updates to support K8s v1.22. Updated default kubernetes version from 1.19 -> 1.21
- Default 'kubernetes_version' updated to 1.21.7
- All example files update to reflect the same

### Test Coverage:
- Verified on Azure with default K8s v1.21.7, viya4 deployment, logging and monitoring were successful.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ ARG AZURECLI_VERSION=2.24.2
 
 FROM hashicorp/terraform:$TERRAFORM_VERSION as terraform
 FROM mcr.microsoft.com/azure-cli:$AZURECLI_VERSION
-ARG KUBECTL_VERSION=1.19.13
+ARG KUBECTL_VERSION=1.21.7
 
 WORKDIR /viya4-iac-azure
 
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ Access to an **Azure Subscription** and an [**Identity**](./docs/user/TerraformA
 
 #### Terraform Requirements:
 - [Terraform](https://www.terraform.io/downloads.html) - v1.0.0
-- [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl) - v1.19.13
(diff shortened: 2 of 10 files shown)


thpang (approved):
These all look good.

thpang:
@riragh I see this is labeled as kubernetes v1.22 support, but the PR is really to update the version to v1.21.7 as the new default. The only support needed for v1.22 would be to stand up a cluster and verify that SAS Viya does indeed start. Knowing you'll also need the updates for ingress-nginx 1.1.0. Is this issue to address the v1.22 issue? There is really no code changes here, just a validation that it will work. But the update on the default value is key here. Just need clarification. Thx

dhoucgitter (commented):
Looks good.

riragh:
> @riragh I see this is labeled as kubernetes v1.22 support, but the PR is really to update the version to v1.21.7 as the new default. The only support needed for v1.22 would be to stand up a cluster and verify that SAS Viya does indeed start. Knowing you'll also need the updates for ingress-nginx 1.1.0. Is this issue to address the v1.22 issue? There is really no code changes here, just a validation that it will work. But the update on the default value is key here. Just need clarification. Thx

Updating the default values is the key here. This change isn't specifically trying to solve any v1.22 issue. Just updating the kubectl and default kubernetes version to v1.21 which gives us 1.20 - 1.22 coverage.

thpang:
Nice. Also update the Dockerfile in this repo as well @riragh This will need it's version modified and then tested.

riragh:
> Nice. Also update the Dockerfile in this repo as well @riragh This will need it's version modified and then tested.

I have the Dockerfile already updated, is there some other docker file that needs update?

thpang:
Nope, just missed it when looking over items. Thanks. All is good this one is ready.

sayeun:
Everything looks good.  +2 👍

PR #224, opened 2022-02-26
(IAC-449) Fix location to drop space else deployment fails

ranieuwe:
There is an error in the variables.tf with the default location (=Azure region). When not specifying a location file and using the default your terraform apply will fail like this:
```
│ Error: "resource_group_name" may only contain alphanumeric characters, dash, underscores, parentheses and periods
│
│   with module.aks.data.azurerm_public_ip.cluster_public_ip[0],
│   on modules/azure_aks/main.tf line 111, in data "azurerm_public_ip" "cluster_public_ip":
│  111:  data "azurerm_public_ip" "cluster_public_ip" {
```
The cause of that is on line https://github.com/sassoftware/viya4-iac-azure/blob/main/modules/azure_aks/main.tf#L116 where it says `resource_group_name = "MC_${var.aks_cluster_rg}_${var.aks_cluster_name}_${var.aks_cluster_location}"`

The `var.aks_cluster_location` is built from the location in the variables.tf, what happens is that the resource group name gets a space in it, MC_name-rg_name_East US, which is not allowed. Correct region would be "eastus", which this PR fixes.

### DIFF
--- a/docs/CONFIG-VARS.md
+++ b/docs/CONFIG-VARS.md
@@ -32,7 +32,7 @@ Terraform input variables can be set in the following ways:
 | Name | Description | Type | Default | Notes |
 | :--- | ---: | ---: | ---: | ---: |
 | prefix | A prefix used in the name of all the Azure resources created by this script. | string | | The prefix string must start with a lowercase letter and contain only lowercase alphanumeric characters and dashes (-), but it cannot end with a dash. |
-| location | The Azure Region to provision all resources in this script. | string | "East US" | |
+| location | The Azure Region to provision all resources in this script. | string | "eastus" | |
 
 ### Azure Authentication
 
--- a/modules/azure_aks/variables.tf
+++ b/modules/azure_aks/variables.tf
@@ -6,7 +6,7 @@ variable aks_cluster_dns_prefix {}
 
 variable "aks_cluster_location" {
   description = "The Azure Region in which all resources in this example should be provisioned"
-  default     = "East US"
+  default     = "eastus"
 }
 
 variable "aks_private_cluster" {
--- a/variables.tf
+++ b/variables.tf
@@ -38,7 +38,7 @@ variable "prefix" {
 }
 variable "location" {
   description = "The Azure Region to provision all resources in this script"
-  default     = "East US"
+  default     = "eastus"
 }
 
 variable "ssh_public_key" {


ranieuwe:
Signed-off-by: Roeland Nieuwenhuis<ronieuwe@microsoft.com>

thpang:
Should be reviewed and tested. @ranieuwe thx for the PR.

riragh:
@ranieuwe, for the error you mentioned above, did you not set the required argument "location" in the terraform.tfvars file?
From what I saw the default location "East US" does get converted to "eastus" correctly. My guess is something else might be causing this issue. To make sure could you let us know what steps let you to this error?

ranieuwe:
No, I have not set any vars whatsoever. This is a vanilla `git clone`, `tf init` `tf plan`, `tf apply`. I am not using any terraform.tfvars on purpose. The logic asks for tenant_id, subscription_id and prefix, that's all. That's what I provided and then it runs into the problem above. 

Error can be replicated by cloning the git repo, `tf init` and then: `terraform apply -destroy -var="tenant_id=72c9d080-bc8c-4c9d-b922-8faa6f3975a8" -var="subscription_id=1d8655d1-6f0f-442a-94b1-1de8b77d2310" -var="prefix=sv4"`

The problem exists because you try to construct the resource group from the location and there is a mismatch. I didn't see any logic that would convert East US to "eastus" in your TF templates anywhere FWIW.

Noticed that I did miss the default configuration setting in the docs, added that to the PR just now.

thpang:
So the usage of location in creating the resource group name is the issue and not the fact it's listed as `East US`. For terraform that is a valid value; however as shown in the error above the generation of the name is where the failure is taking place.

The code in question is here : https://github.com/sassoftware/viya4-iac-azure/blob/main/modules/azure_aks/main.tf#L116

@riragh, it may be better to either setup the values without spaces or have a local variable that would do the conversion if we allow this.

@ranieuwe if you could please review the docs here : https://github.com/sassoftware/viya4-iac-azure/blob/main/docs/CONFIG-VARS.md#required-variables that outlines which variables are needed and wihch can be optional. Thanks.

ranieuwe:
> So the usage of location in creating the resource group name is the issue and not the fact it's listed as East US. For terraform that is a valid value; however as shown in the error above the generation of the name is where the failure is taking place.

Yes. The PR is to correct the default for the required variable, as it isn't working.

thpang:
Yes the PR is correct, I am trying to make sure the IAC team understands what is happening and giving them some guidance here. The documentation does say the `location` variable is required which you failed to provide. Someone will be looking at this in the near future. Thx.

riragh:
Thanks for clarification @ranieuwe. I made another small update to location field present under modules/azure_aks/variables.tf.
I have tested the changes in this PR, good to be merged after approval.

thpang (approved):
Looks good.

jarpat (approved):
👍

PR #227, opened 2022-04-05
(IAC-374) Update BYO Subnet Documentation

jarpat:
## Changes
* Updates the BYO Subnet documentation to make it clear that if the user opts for an external Postgres instance they will need to define service endpoints in their subnets.
* Minor doc formatting/cleanup


thpang (approved):
LGTM

PR #234, opened 2022-04-27
(IAC-451) Create initial sample input file for SingleStore Azure cluster

dhoucgitter:
Initial draft of best practice terraform inputs for the creation of SingleStore compatible Azure cluster for a Viya4 deployment.

### DIFF
--- a/examples/sample-input-singlestore.tfvars
+++ b/examples/sample-input-singlestore.tfvars
@@ -1,0 +1,152 @@
+# !NOTE! - These are only a subset of CONFIG-VARS.md provided as examples.
+# Customize this file to add any variables from 'CONFIG-VARS.md' whose default values
+# you want to change.
+
+# ****************  REQUIRED VARIABLES  ****************
+# These required variables' values MUST be provided by the User
+prefix   = "<prefix-value>" # this is a prefix that you assign for the resources to be created
+location = "<azure-location-value>" # e.g., "eastus2"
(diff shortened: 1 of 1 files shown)


examples/sample-input-singlestore.tfvars
```diff
@@ -0,0 +1,152 @@
...
+    "machine_type" = "Standard_E8ds_v5"         
+    "os_disk_size" = 200
+    "min_nodes" = 0
+    "max_nodes" = 7
+    "max_pods" = 110
+    "node_taints" = ["workload.sas.com/class=singlestore:NoSchedule"]
+    "node_labels" = {
+      "workload.sas.com/class" = "singlestore"
+    }
+  },
```
thpang, line 54: I would suggest adding this at the end given the issues we've seen with AWS when/if we need to support ss there. I'd also remove the extra space.

thpang (changes requested):
Comments for change.

examples/sample-input-singlestore.tfvars
dhoucgitter: point taken, thanks for the suggestion

thpang (approved):

dhoucgitter:
@thpang deleting this branch per request and replacing it with a branch named singlestore which I will seed with udpated sample-input-singlestore.tfvars file.

PR #239, opened 2022-06-23
(IAC-506) Added if-else logic to avoid duplicate exports

riragh:
The cloud-init assumes the aks and misc subnets will be different but that's not always the case. The cloud-config should check to ensure that duplicate lines don't end up in /etc/exports.

## Changes
Added condition to avoid duplicate exports.

## Tests
- Default cluster creation successful
- BYO cluster creation

** Additionally user verified this in their own testing scenario.

### DIFF
--- a/files/cloud-init/nfs/cloud-config
+++ b/files/cloud-init/nfs/cloud-config
@@ -56,8 +56,9 @@ runcmd:
   #
   # Update /etc/exports - NOTE: The CIDR provided works for the whole VPC
   #
-  - echo "/export         ${aks_cidr_block}(rw,no_root_squash,async,insecure,fsid=0,crossmnt,no_subtree_check)" >> /etc/exports
-  - echo "/export         ${misc_cidr_block}(rw,no_root_squash,async,insecure,fsid=0,crossmnt,no_subtree_check)" >> /etc/exports
+  - if [ ${aks_cidr_block} == ${misc_cidr_block} ]; then echo "/export         ${aks_cidr_block}(rw,no_root_squash,async,insecure,fsid=0,crossmnt,no_subtree_check)" >> /etc/exports; else echo "/export         ${aks_cidr_block}(rw,no_root_squash,async,insecure,fsid=0,crossmnt,no_subtree_check)" >> /etc/exports && echo "/export         ${misc_cidr_block}(rw,no_root_squash,async,insecure,fsid=0,crossmnt,no_subtree_check)" >> /etc/exports; fi
+  # - echo "/export         ${aks_cidr_block}(rw,no_root_squash,async,insecure,fsid=0,crossmnt,no_subtree_check)" >> /etc/exports
+  # - echo "/export         ${misc_cidr_block}(rw,no_root_squash,async,insecure,fsid=0,crossmnt,no_subtree_check)" >> /etc/exports
   #
   # Restart nfs-server service
   #


files/cloud-init/nfs/cloud-config
```diff
@@ -56,8 +56,9 @@ runcmd:
   #
   # Update /etc/exports - NOTE: The CIDR provided works for the whole VPC
   #
-  - echo "/export         ${aks_cidr_block}(rw,no_root_squash,async,insecure,fsid=0,crossmnt,no_subtree_check)" >> /etc/exports
-  - echo "/export         ${misc_cidr_block}(rw,no_root_squash,async,insecure,fsid=0,crossmnt,no_subtree_check)" >> /etc/exports
+  - if [ ${aks_cidr_block} == ${misc_cidr_block} ]; then echo "/export         ${aks_cidr_block}(rw,no_root_squash,async,insecure,fsid=0,crossmnt,no_subtree_check)" >> /etc/exports; else echo "/export         ${aks_cidr_block}(rw,no_root_squash,async,insecure,fsid=0,crossmnt,no_subtree_check)" >> /etc/exports && echo "/export         ${misc_cidr_block}(rw,no_root_squash,async,insecure,fsid=0,crossmnt,no_subtree_check)" >> /etc/exports; fi
+  # - echo "/export         ${aks_cidr_block}(rw,no_root_squash,async,insecure,fsid=0,crossmnt,no_subtree_check)" >> /etc/exports
+  # - echo "/export         ${misc_cidr_block}(rw,no_root_squash,async,insecure,fsid=0,crossmnt,no_subtree_check)" >> /etc/exports
```
thpang, line 61: The code snipped from the AWS solution seems a bit cleaner as it combines the cidr_blocks and then combines those. If they are the same, then only 1 entry:

```
  #
  # Update /etc/exports
  #
  - for cidr_block in ${public_subnet_cidrs} ${private_subnet_cidrs}
  - do
  -   echo "/export         $cidr_block(rw,no_root_squash,async,insecure,fsid=0,crossmnt,no_subtree_check)" >> /etc/exports
  - done
```
The code example can be found [here](https://github.com/sassoftware/viya4-iac-aws/blob/main/files/cloud-init/nfs/cloud-config#L56-L62)

thpang (changes requested):
Please review the attached example from one of our other cloud provider solutions.

files/cloud-init/nfs/cloud-config
```diff
@@ -56,8 +56,13 @@ runcmd:
...
   #
-  - echo "/export         ${aks_cidr_block}(rw,no_root_squash,async,insecure,fsid=0,crossmnt,no_subtree_check)" >> /etc/exports
-  - echo "/export         ${misc_cidr_block}(rw,no_root_squash,async,insecure,fsid=0,crossmnt,no_subtree_check)" >> /etc/exports
+  - if [ ${aks_cidr_block} == ${misc_cidr_block} ]; 
+  - then 
+  -   echo "/export         ${aks_cidr_block}(rw,no_root_squash,async,insecure,fsid=0,crossmnt,no_subtree_check)" >> /etc/exports;
+  - else
+  -   echo "/export         ${aks_cidr_block}(rw,no_root_squash,async,insecure,fsid=0,crossmnt,no_subtree_check)" >> /etc/exports && \
+  -   echo "/export         ${misc_cidr_block}(rw,no_root_squash,async,insecure,fsid=0,crossmnt,no_subtree_check)" >> /etc/exports;
+  - fi
```
thpang, line 65: Would the for loop example not work here? Was wondering if that could be implemented as we have this on the AWS side. I have not checked on the GCP side. Was thinking that the combined items since they were the same would only yield 1 entry. Can you test this?

thpang (commented):
Looking for verification and some testing

files/cloud-init/nfs/cloud-config
riragh: Not sure how to test the whole scenario but I tried testing the for loop logic separately. For loop doesn't seem to combine the items, treats it as a list and iterates over all the items. We would need some extra logic to make it a unique list for the 'for' loop to iterate over.
Tbh if-else seems much simpler to accomplish this task.

thpang: Just would need to test when both networks are the same. If this is no feasible, then the code base works fine with the updated if statement.

thpang (approved):
With the new if logic this looks good.

riragh:
Still waiting for confirmation from the issue reporter to see this resolves the issue in their testing scenario. I'll keep this PR open, in case we need further updates.

riragh:
Switching around the if condition seems to work better for Stephen.

thpang:
Not sure why were are hurrying this one along. The code keeps getting pushed and I am not seeing any signs of actual thought or testing here. Is there a reason?  What/Why does the code work better for Stephen? The code should be robust and easy to read along with functional. Please take the time to clean this one up and then once tested ask for a review. If I am missing something let me know.

PR #240, opened 2022-06-23
(IAC-506) Modified if-else logic to avoid duplicate exports

stturc:


riragh (approved):
LGTM

riragh:
Switching around the if condition seems to work better for Stephen. Merging this PR into the parent PR #239.

PR #249, opened 2022-08-16
implement sku_tier for IAC-639

ajeffowens:
This adds sku_tier to variables.tf.  Choosing "Free" (the default) results in a cluster "optimized for availability".  Choosing "Paid" optimizes for availability (and cost more $).  Default behavior does not change.  Tested and working


ajeffowens:
tflint caught a typo.  will get that cleaned up.

PR #250, opened 2022-08-16
implement sku_tier for IAC-639

ajeffowens:
This adds sku_tier to variables.tf. Choosing "Free" (the default) results in a cluster "optimized for availability". Choosing "Paid" optimizes for availability (and cost more $). Default behavior does not change. Tested and working


thpang (commented):
@ajeffowens, add this variable to the CONFIG-VARS.md file to complete the change. Is this only an AKS item? If so I'd suggest changing the name at a higher level to `aks_sku_tier`. This will make it clear to the end user and it'll be in the `General` section of the doc.

thpang (changes requested):
LGTM. Base changes need merging into Fork for acceptance.

PR #254, opened 2022-08-23
feat: (IAC-583): AKS add support for K8s 1.24

riragh:
# Background:
Starting from Kubernetes version `1.24.0` service account token is not automatically generated, thus it has to be created separately. The following resources were updated by the provider `hashicorp/kubernetes` to handle this change in `v2.13.0`: `d/kubernetes_service_account`, `r/kubernetes_default_service_account`, `r/kubernetes_service_account`. For Kubernetes clusters running v1.24+ `default_secret_name` will be empty. A warning message will be printed once any of the above resources are in use. 

# Changes:
- To support 1.24+ Kubernetes version in AKS, secret is manually created for the service account
- The default K8s value is updated from 1.22.6 to 1.23.8
- Minor warning fix for provider `hashicorp/azuread`

# Tests:
1. Verified cluster creation with K8s version: 1.24.0 was successful. Viya deployment was successful and application was accessible
2. Verified cluster was created successfully with the new default K8s version: 1.23.8. Viya deployment was successful and application was accessible
3. Verified cluster creation with K8s version: 1.22.6 was successful. Viya deployment was successful and application was accessible
4. Verified cluster creation with version 1.24.0 and create_static_kubeconfig=false was successful. Provider based kubeconfig was generated and used to deploy Viya successfully.

### DIFF
--- a/modules/kubeconfig/main.tf
+++ b/modules/kubeconfig/main.tf
@@ -21,6 +21,7 @@ data "template_file" "kubeconfig_provider" {
 
 # 1.24 change: Create service account secret
 resource "kubernetes_secret" "sa_secret" {
+  count = var.create_static_kubeconfig ? 1 : 0
   metadata {
     name      = local.service_account_secret_name
     namespace = var.namespace
@@ -36,7 +37,7 @@ resource "kubernetes_secret" "sa_secret" {
 data "kubernetes_secret" "sa_secret" {
   count = var.create_static_kubeconfig ? 1 : 0
   metadata {
-    name = "${kubernetes_secret.sa_secret.metadata.0.name}"
+    name = "${kubernetes_secret.sa_secret[0].metadata.0.name}"
     namespace = var.namespace
   }
 }
@@ -55,8 +56,8 @@ data "template_file" "kubeconfig_sa" {
   }
 }
 
-## Warning message from hashicorp/terraform-provider-kubernetes is expected 
-# Message: "Warning: 'default_secret_name' is no longer applicable for Kubernetes 'v1.24.0' and above"
+# Starting K8s v1.24+ hashicorp/terraform-provider-kubernetes issues warning message:
+# "Warning: 'default_secret_name' is no longer applicable for Kubernetes 'v1.24.0' and above"
 resource "kubernetes_service_account" "kubernetes_sa" {
   count = var.create_static_kubeconfig ? 1 : 0
   metadata {
@@ -64,7 +65,6 @@ resource "kubernetes_service_account" "kubernetes_sa" {
     namespace = var.namespace
   }
 }
-
 
 resource "kubernetes_cluster_role_binding" "kubernetes_crb" {
   count = var.create_static_kubeconfig ? 1 : 0
--- a/modules/kubeconfig/main.tf
+++ b/modules/kubeconfig/main.tf
@@ -20,20 +20,6 @@ data "template_file" "kubeconfig_provider" {
 }
 
 # Service Account based kube config data/template/resources
-# 1.24 change: Create service account secret
-resource "kubernetes_secret" "sa_secret" {
-  count = var.create_static_kubeconfig ? 1 : 0
-  metadata {
-    name      = local.service_account_secret_name
-    namespace = var.namespace
-    annotations = {
-      "kubernetes.io/service-account.name" = local.service_account_name
-    }
-  }
-
-  type = "kubernetes.io/service-account-token"
-}
-
 data "template_file" "kubeconfig_sa" {
   count = var.create_static_kubeconfig ? 1 : 0
   template = file("${path.module}/templates/kubeconfig-sa.tmpl")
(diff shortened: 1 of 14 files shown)


modules/kubeconfig/main.tf
```diff
@@ -19,11 +19,24 @@ data "template_file" "kubeconfig_provider" {
   }
 }
 
+# 1.24 change: Create service account secret
+resource "kubernetes_secret" "sa_secret" {
+  metadata {
```
thpang, line 24: There should be a guard on this code base. Always list the data element before the resource. You'll need a guard like this:

`count = var.create_stataic_kubeconfig ? 0 : 1` 

for this block.

We either use the data source or create our own resource.

modules/kubeconfig/main.tf
```diff
@@ -19,11 +19,24 @@ data "template_file" "kubeconfig_provider" {
...
+
+  type = "kubernetes.io/service-account-token"
+}
+
 # Service Account based kube config data/template/resources
 data "kubernetes_secret" "sa_secret" {
   count = var.create_static_kubeconfig ? 1 : 0
   metadata {
-    name      = kubernetes_service_account.kubernetes_sa.0.default_secret_name
+    name = "${kubernetes_secret.sa_secret.metadata.0.name}"
```
thpang, line 39: If this is the new secret name, what's the value of this string? It's metadata so it's just passed in so k8s can look it up.

modules/kubeconfig/main.tf
```diff
@@ -19,11 +19,24 @@ data "template_file" "kubeconfig_provider" {
...
+    namespace = var.namespace
+    annotations = {
+      "kubernetes.io/service-account.name" = local.service_account_name
+    }
+  }
+
+  type = "kubernetes.io/service-account-token"
+}
+
 # Service Account based kube config data/template/resources
```
thpang, line 35: Is this comment still true?

modules/kubeconfig/main.tf
```diff
@@ -50,6 +65,7 @@ resource "kubernetes_service_account" "kubernetes_sa" {
   }
 }
 
+
```
thpang, line 68: extra line remove

thpang (changes requested):

modules/kubeconfig/main.tf
riragh: This comment is true because the Service Account is still based off the template. The secret for the service account is the new resource created.

modules/kubeconfig/main.tf
riragh: I'll add that.

modules/kubeconfig/main.tf
riragh: Value of this string will be determined after apply, but it's basically pulling value from line #26.

modules/kubeconfig/main.tf
riragh: Done

modules/kubeconfig/main.tf
riragh: Added in the requested change.

modules/kubeconfig/main.tf
```diff
@@ -19,11 +19,25 @@ data "template_file" "kubeconfig_provider" {
...
+  type = "kubernetes.io/service-account-token"
+}
+
 # Service Account based kube config data/template/resources
 data "kubernetes_secret" "sa_secret" {
   count = var.create_static_kubeconfig ? 1 : 0
   metadata {
-    name      = kubernetes_service_account.kubernetes_sa.0.default_secret_name
+    name = "${kubernetes_secret.sa_secret[0].metadata.0.name}"
```
thpang, line 40: Why variable syntax here?

thpang (changes requested):

modules/kubeconfig/main.tf
riragh: The logic above is correct. The data source and the resource for secret are not either or. Previously the data was reading the secret from the service account's default secret however starting K8s v1.24.0 default secret is no longer being generated for the service account and hence the data can't get any secret. So the solution to this was to create a new resource kubernetes_secret which provides the secret to the data source.
Discussed this with Thomas, the usage of data/resource pair and the referencing syntax in the current code is different. In the latest change I have updated the logic to only reference the resource and removed the data source for the secret.

modules/kubeconfig/main.tf
riragh: data source is removed in the latest change. Code now only refers the new resource for the secret.

jarpat (approved):

modules/kubeconfig/main.tf
```diff
@@ -36,12 +28,28 @@ data "template_file" "kubeconfig_sa" {
...
-    token        = lookup(data.kubernetes_secret.sa_secret.0.data,"token", "")
+    ca_crt       = base64encode(lookup(kubernetes_secret.sa_secret.0.data,"ca.crt", ""))
+    token        = lookup(kubernetes_secret.sa_secret.0.data,"token", "")
     namespace    = var.namespace
   }
 }
 
+# 1.24 change: Create service account secret
+resource "kubernetes_secret" "sa_secret" {
+  count = var.create_static_kubeconfig ? 1 : 0
```
thpang, line 39: thanks for adjusting this one. But now What if the user does not request a static kube config file and its generated by the provider and that provider is using 1.24. Does that cause issues in the cluster? or do we still need to create a default secret for the cluster? Just verifying.

riragh: Looking at the code for provider based kubeconfig the values are added using the data source element at line #8 with the variables defined on line #15-#18. Value of these variables is probably generated here: https://github.com/sassoftware/viya4-iac-azure/blob/main/main.tf#L169
This is independent of the service_account secret.

thpang: I understand we don't create the secret, but does the provider create a default secret the cluster needs if our code is not being used to generate the kubeconfig file? Wondering if the SAS deployment is dependent on this and if there's a need for a a default secret in the entire cluster. The issue I believe is larger than the scope of what might be captured here.

riragh:
Updates made to unify the code across cloud providers. Verified cluster creation with both static kubeconfig and provider based kubeconfig. Changes ready for review.

thpang (approved):
LGTM

PR #255, opened 2022-08-24
Add variable to control the private_dns_zone_id value

AWSmith0216:
For our internal VPNed subscriptions, the use of the hardcoded **System** value for **private_dns_zone_id** does not seem to work. The worker nodes fail to contact the API server with an error similar to:

`server can't find adsmit-foo-aks-38c762d3.48995284-7e91-4ff1-bd6f-2ea6fbf3ae54.privatelink.eastus.azmk8s.io: NXDOMAIN
`

As for why the error occurs, I'm not sure. When set to **System**, a private DNS zone is created with the appropriate A record that points the private FQDN to the internal IP. The DNS zone link to the vnet is also setup. Why this link doesn't seem to be working as expected, I don't know.

The VNETs in our CIS provided VPNed subscriptions have custom DNS servers set, rather than using the **Default (Azure-provided)** option. I'm assuming that's related to the problem. According to the [terraform doc](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster), setting this value to **None** means that you're bringing your own DNS servers. When set to **None**, no private DNS zone is created, there's apparently no private FQDN, and the public FQDN resolves to the private IP from within the VNET.

I suppose why **System** doesn't work is somewhat irrelevant though. This pull request just makes it something the user can adjust as they wish, rather than having it hardcoded. I also purposely didn't add this variable to the BYO Networking README so it would be considered "use at your own risk".

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -154,6 +154,7 @@ module "aks" {
   aks_network_plugin                       = var.aks_network_plugin
   aks_network_policy                       = var.aks_network_policy
   aks_dns_service_ip                       = var.aks_dns_service_ip
+  aks_private_dns_zone_id                  = var.aks_private_dns_zone_id
   aks_docker_bridge_cidr                   = var.aks_docker_bridge_cidr
   cluster_egress_type                      = local.cluster_egress_type
   aks_pod_cidr                             = var.aks_pod_cidr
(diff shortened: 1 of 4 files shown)


AWSmith0216:
Any chance we could get this merged into the repo?

AWSmith0216:
Looks like the other PR would accomplish the same goal, so I'm fine with closing this one.

PR #257, opened 2022-09-06
feat: (IAC-403): Update AzureRM Terraform provider version

riragh:
# Changes:
Upgrade `azurerm` provider version from 2.62.0 to 3.21.1. Followed the upgrade document [3.0-upgrade-guide](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/guides/3.0-upgrade-guide)

# Tests:
Following Scenarios were verified. Cluster creation was successful for all the scenarios. See internal ticket for details.
1. All defaults
2. Azure monitor enabled: create_aks_azure_monitor = true
3. create_container_registry = true, container_registry_sku = "Premium" and container_registry_geo_replica_locs = ["eastus", "southcentralus"]
4. create_container_registry = true, container_registry_sku = "Premium", container_registry_geo_replica_locs = null
5. default_nodepool_availability_zones = ["1","2"]

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -101,21 +101,13 @@ resource "azurerm_container_registry" "acr" {
   # for backwards compatability.
   #
   dynamic "georeplications" {
-    for_each = (local.container_registry_sku == "Premium" && var.container_registry_geo_replica_locs != null) ? var.container_registry_geo_replica_locs : (local.container_registry_sku == "Premium" ? [] : null)
+    for_each = (local.container_registry_sku == "Premium" && var.container_registry_geo_replica_locs != null) ? toset(
+      var.container_registry_geo_replica_locs) : []
     content {
-      location                = var.container_registry_geo_replica_locs.value.location
+      location                = georeplications.key
       tags                    = var.tags
     }
   }
-
-#   georeplications = (local.container_registry_sku == "Premium" && var.container_registry_geo_replica_locs != null) ? [
-#     for location_item in var.container_registry_geo_replica_locs:
-#       {
-#         location = location_item
-#         tags     = var.tags
-#       }
-#   ] : local.container_registry_sku == "Premium" ? [] : null
-
   tags                     = var.tags
 }
 
--- a/main.tf
+++ b/main.tf
@@ -100,13 +100,21 @@ resource "azurerm_container_registry" "acr" {
   # Moving from deprecated argument, georeplication_locations, but keeping container_registry_geo_replica_locs
   # for backwards compatability.
   #
-  georeplications = (local.container_registry_sku == "Premium" && var.container_registry_geo_replica_locs != null) ? [
-    for location_item in var.container_registry_geo_replica_locs:
-      {
-        location = location_item
-        tags     = var.tags
-      }
-  ] : local.container_registry_sku == "Premium" ? [] : null
+  dynamic "georeplications" {
+    for_each = (local.container_registry_sku == "Premium" && var.container_registry_geo_replica_locs != null) ? var.container_registry_geo_replica_locs : (local.container_registry_sku == "Premium" ? [] : null)
+    content {
+      location                = var.container_registry_geo_replica_locs.value.location
+      tags                    = var.tags
+    }
+  }
+
+#   georeplications = (local.container_registry_sku == "Premium" && var.container_registry_geo_replica_locs != null) ? [
+#     for location_item in var.container_registry_geo_replica_locs:
+#       {
+#         location = location_item
+#         tags     = var.tags
+#       }
+#   ] : local.container_registry_sku == "Premium" ? [] : null
 
   tags                     = var.tags
 }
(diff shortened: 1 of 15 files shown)


main.tf
```diff
@@ -100,14 +100,14 @@ resource "azurerm_container_registry" "acr" {
...
-  ] : local.container_registry_sku == "Premium" ? [] : null
-
+  dynamic "georeplications" {
+    for_each = (local.container_registry_sku == "Premium" && var.container_registry_geo_replica_locs != null) ? toset(
+      var.container_registry_geo_replica_locs) : []
+    content {
+      location                = georeplications.key
+      tags                    = var.tags
+    }
+  }
```
thpang, line 110: Was there a need to change this to a dynamic block? Asking in general as the new module also treats the `georeplications` parameter as optional and could take a value of `null`. Not saying the dynamic block is a bad thing, just wondering why we spent the time to re-write the code here.

modules/aks_node_pool/main.tf
```diff
@@ -5,7 +5,7 @@ resource "azurerm_kubernetes_cluster_node_pool" "autoscale_node_pool" {
   name                         = var.node_pool_name
   kubernetes_cluster_id        = var.aks_cluster_id
   vnet_subnet_id               = var.vnet_subnet_id
-  availability_zones           = var.availability_zones
+  zones                        = var.zones
```
thpang, line 8: I know they've changed the `zone` type naming. But wanted to be sure there were other breaking changes in this update outside of us changing our variables that the user may have had as inputs into the scripts. Is this the case?

versions.tf
```diff
@@ -5,7 +5,7 @@ terraform {
   required_providers {
     azurerm = {
       source  = "hashicorp/azurerm"
-      version = "2.62.0"
+      version = "3.20.0"
```
thpang, line 8: They're up to 3.21.1 now. Let's make sure we can support close to the latest version as we keep moving forward.

thpang (commented):
Just a few comments on the work done so far.

modules/aks_node_pool/main.tf
riragh: In this upgrade there were a couple variables/blocks that were deprecated, few were renamed and few had change of type
| Original  | New  | Update  | Affects user input |
|---|---|---|---|
| availability_zones | zones | Name change only | Yes |
| nfs_raid_disk_zones |  nfs_raid_disk_zone | Instead of collection accepts only a single string value  | Yes |
| enforce_private_link_endpoint_network_policies | private_endpoint_network_policies_enabled | Name change only | Yes |
| enforce_private_link_service_network_policies  | private_link_service_network_policies_enabled | Name change only | Yes |
| georeplications | georeplications| property needs to be defined as block | No |
| role_based_access_control | role_based_access_control_enabled | block deprecated in favour of the new property | No|
| addon_profile | NA | block removed in favour of other properties | No |
| addon_profile.oms_agent | oms_agent | property updated as individual block | No |
| addon_profile.http_application_routing_enabled | http_application_routing_enabled | individual property | No |
| addon_profile.kube_dashboard | NA | Property removed | No |

main.tf
riragh: The argument "georeplications" is updated to be defined as a block of type. You are right, the new module does treat `georeplications` as an optional argument and can accept null to ignore the block from being added.
The original code had logic in place for `container_registry_sku = "Premium" and container_registry_geo_replica_locs is set` so updated the existing property to a dynamic block to support the case were user provides multiple geo replication locations. And setting empty `[]`, omits the property

jarpat:
Minor doc update that can also be included with this change:
https://github.com/sassoftware/viya4-iac-azure/blob/IAC-403/README.md?plain=1#L121
The link to the azurerm kubernetes cluster doc is currently invalid, just need to remove the '.html' part from the end

riragh:
> Minor doc update that can also be included with this change: https://github.com/sassoftware/viya4-iac-azure/blob/IAC-403/README.md?plain=1#L121 The link to the azurerm kubernetes cluster doc is currently invalid, just need to remove the '.html' part from the end

Done

jarpat (approved):

dhoucgitter (approved):

modules/aks_node_pool/main.tf
thpang: Ok, so several updates to existing user input correct? This is definitely a breaking change.

main.tf
thpang: Sounds good.

outputs.tf
```diff
@@ -13,7 +14,7 @@ output "kube_config" {
 }
 
 output "aks_cluster_node_username" {
-  value = module.aks.cluster_username
+  value = nonsensitive(module.aks.cluster_username)
```
thpang, line 17: And now why are we actually marking this one as nonsensitive. Seems like the consistency on items is a bit off. If needed converse with the group about these items.

thpang (changes requested):

outputs.tf
```diff
@@ -1,6 +1,7 @@
 # #aks
 output "aks_host" {
   value = module.aks.host
+  sensitive = true
```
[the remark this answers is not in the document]
riragh: Yes, cluster creation fails with `Error: Output refers to sensitive values`. So added sensitive = true.

outputs.tf
riragh: Same error on this one `Error: Output refers to sensitive values`. I wasn't sure if wanted to mask the username from the log output so added as nonsensitive to keep it as before.
Do you want me to change to sensitive for consistency?

outputs.tf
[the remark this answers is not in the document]
thpang: Ok, sounds good.

outputs.tf
thpang: Ok, so if that's the reasoning why was the aks_host item made sensitive? Not consistent.

thpang (approved):
With the latest updates. Approving this one. Thanks again.

outputs.tf
riragh: Yes, should have kept them consistent from the beginning. I updated this in the latest change for all to be consistent.

PR #263, opened 2022-10-11
feat: (IAC-643): Enable specification of azurerm_netapp_volume.anf.network_features

riragh:
# Changes:
- New variable `netapp_network_features` added to accept the ANF network_features value. Accepted values are `Basic` or `Standard`, default value is set to `Basic`.

- azurerm version updated to latest `v3.26.0` (Since the last update azurerm jumped a few version), no underlying code changes for this version.

# Tests:
Verified following scenarios, see internal ticket for details:

- Scenario 1: netapp_network_features set to Standard
- Scenario 2: netapp_network_features not set, defaults to Basic 
- Scenario 3: SingleStore enabled with netapp_network_features set to Standard

### DIFF
--- a/modules/azurerm_netapp/variables.tf
+++ b/modules/azurerm_netapp/variables.tf
@@ -23,6 +23,10 @@ variable "subnet_id" {
   description = "Azure subnet id for Azure NetApp Files"
 }
 
+variable "network_features" {
+  description = "Indicates which network feature to use, accepted values are Basic or Standard, it defaults to Basic if not defined."
+}
+
 # https://docs.microsoft.com/en-us/azure/azure-netapp-files/azure-netapp-files-service-levels
 variable "service_level" {
   description = "The target performance of the file system. Valid values include Premium, Standard, or Ultra."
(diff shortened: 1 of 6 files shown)


modules/azurerm_netapp/variables.tf
```diff
@@ -23,6 +23,10 @@ variable "subnet_id" {
   description = "Azure subnet id for Azure NetApp Files"
 }
 
+variable "network_features" {
+  description = "Indicates which network feature to use, accepted values are Basic or Standard, it defaults to Basic if not defined."
+}
```
thpang, line 28: Wow have a lot of non-existent `type` fields in this file. If we don't want to force it to something we can add 
```yaml
  type = null
```
or 
```yaml
  type = any
```
This is just a general comment not for this issue, but need to start making sure we know the types of our variables in the future.

thpang, line 28: So for this code change we do actually say the default is `Basic` so we need at least that defined here

thpang (changes requested):

modules/azurerm_netapp/variables.tf
riragh: Done. Added `type = string` for the new variable. I'll create a separate new ticket to make similar updates to the other old variables.

riragh: Done.

thpang (approved):
LGTM

jarpat (approved):

PR #264, opened 2022-10-15
docs: (IAC-403): documentation updates

riragh:
# Changes:
- Renamed the missed "nfs_raid_disk_zones" to "nfs_raid_disk_zone"
- Removed the reference of "create_aks_azure_monitor" from example sample files, as this is an undocumented feature.

### DIFF
--- a/examples/sample-input-ppg.tfvars
+++ b/examples/sample-input-ppg.tfvars
@@ -108,7 +108,4 @@ nfs_vm_zone          = 1
 
 nfs_raid_disk_size  = 128
 nfs_raid_disk_type  = "Standard_LRS"
-nfs_raid_disk_zones = ["1"]
-
-# Azure Monitor
-create_aks_azure_monitor = false
+nfs_raid_disk_zone = "1"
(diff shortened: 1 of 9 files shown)


examples/sample-input-ppg.tfvars
```diff
@@ -108,7 +108,4 @@ nfs_vm_zone          = 1
 
 nfs_raid_disk_size  = 128
 nfs_raid_disk_type  = "Standard_LRS"
-nfs_raid_disk_zones = ["1"]
-
-# Azure Monitor
-create_aks_azure_monitor = false
+nfs_raid_disk_zone = "1"
```
thpang, line 111: I am guessing somewhere in the docs for this upgrade we document why we went from having disk capable of spanning multiple zones to now only supporting a single zone managed disk. This would be important given our sample file only shows an array of 1 element but there could be customers who have spread disk across multiple zones in their configuration.

thpang (approved):
All looks good. Just need to be sure we document the change from `nfs_raid_disk_zones` -> `nfs_raid_disk_zone` in case any customers used this multizonal option in the past.

examples/sample-input-ppg.tfvars
riragh: This change was made following the [3.0-upgrade-guide](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/guides/3.0-upgrade-guide).

The field zones has been renamed zone (as this only accepts a single value) to be consistent across the Provider. See [Resource: azurerm_managed_disk](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/guides/3.0-upgrade-guide#resource-azurerm_managed_disk).
The release note will include the information on the changed inputs along with the reference links. I also have this information noted in the ticket IAC-403.

jarpat (approved):
👍

PR #268, opened 2022-10-21
feat: (IAC-438) Increase default node count and size for minimal cluster config

dhoucgitter:
### Changes
The current minimally sized cluster sample file does not allow for reliable successful deployment of cluster-logging and Viya and sometimes result in a failed logging deployment with errors indicating that insufficient initial cluster resources are the cause.
Increasing the default node count and default vm size in the current minimally sized cluster sample values to 
```
default_nodepool_min_nodes = 2
default_nodepool_vm_type  = "Standard_D4_v3"
```
remedies the problem for a marginally small increase in cost.

### Tests
Ran three consecutive deployments ("cluster-logging,viya,install") against an Azure cluster starting out each time with only the 2 default nodes sized as indicated above. Each time cluster-logging deployed without errors and all logging namespace pods reached the Running status. Internal ticket contains test output detail.

### DIFF
--- a/examples/sample-input-minimal.tfvars
+++ b/examples/sample-input-minimal.tfvars
@@ -33,9 +33,9 @@ container_registry_admin_enabled    = false
 
 # AKS config
 kubernetes_version         = "1.23.8"
-default_nodepool_min_nodes = 1
+default_nodepool_min_nodes = 2
+default_nodepool_vm_type   = "Standard_D8s_v4"
 #v3 still has local temp storage
-default_nodepool_vm_type   = "Standard_D2_v3"
 
 # AKS Node Pools config - minimal
 cluster_node_pool_mode = "minimal"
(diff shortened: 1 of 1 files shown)


examples/sample-input-minimal.tfvars
```diff
@@ -33,9 +33,9 @@ container_registry_admin_enabled    = false
 
 # AKS config
 kubernetes_version         = "1.23.8"
-default_nodepool_min_nodes = 1
+default_nodepool_min_nodes = 2
+default_nodepool_vm_type   = "Standard_D8s_v4"
```
thpang, line 37: This value does not match the Jira story of this issues vm size. What was the need for the increase? Asking as it now affects the customer by have 2x the machines which translates to 8x the cost of the default node pool which typically has minimal impact on cost.

dhoucgitter: @thpang Good catch, I entered the wrong value for vm_type on Friday, the value that I tested with is the value reflected in this PR Changes section and the Jira story, I've corrected it in the last commit.

thpang (approved):

riragh (approved):

PR #270, opened 2022-11-11
feat: (IAC-793): support multiple worker/agent availability zones

jason-matthew:
# Feature:
- Allow IaC users to specify a list of availability zones for AKS worker node pools

### Summary:
- Feature introduces parity between system and worker node pool. Both pools now support a user-defined list of availability zones
- Change introduces var `node_pools_availability_zones` while preserving singular variant `node_pools_availability_zone`
- Existing variable will be honored and given priority
- Need to consider eventual deprecation of `node_pools_availability_zone`

# Tests:
Verified the following scenarios:
|:----|:----|:----|:----|
|1|Create cluster with new variable node_pools_availability_zones as list|Fast 2020|Cluster creation with multi-zone setup was successful. Viya4 deployment on the aks cluster stabilized and applications are accessible|
|2|Create cluster with old variable node_pools_availability_zone as string|Fast 2020|Cluster creation with single availability zone was succesful. Viya4 deployment stabilized and applications are accessible|

### DIFF
--- a/docs/CONFIG-VARS.md
+++ b/docs/CONFIG-VARS.md
@@ -63,7 +63,7 @@ To do set these permissions as part of this Terraform script, specify ranges of 
 
 NOTE: When deploying infrastructure into a private network (e.g. a VPN), with no public endpoints, the options documented in this block are not applicable.
 
-NOTE: The script will either create a new NSG, or use an existing NSG, if specified in the [`nsg_name`](#use-existing) variable. 
+NOTE: The script will either create a new NSG, or use an existing NSG, if specified in the [`nsg_name`](#use-existing) variable.
 
 You can use `default_public_access_cidrs` to set a default range for all created resources. To set different ranges for other resources, define the appropriate variable. Use an empty list `[]` to disallow access explicitly.
 
@@ -75,7 +75,7 @@ You can use `default_public_access_cidrs` to set a default range for all created
 | postgres_public_access_cidrs | IP address ranges allowed to access the Azure PostgreSQL Flexible Server | list of strings || Opens port 5432 by adding Ingress Rule on the NSG. Only used when creating postgres instances. |
 | acr_public_access_cidrs | IP address ranges allowed to access the ACR instance | list of strings || Only used with `create_container_registry=true` |
 
-**NOTE:** In a SCIM environment, the AzureActiveDirectory service tag must be granted access to port 443/HTTPS for the Ingress IP address. 
+**NOTE:** In a SCIM environment, the AzureActiveDirectory service tag must be granted access to port 443/HTTPS for the Ingress IP address.
 
 ## Networking
 
@@ -147,7 +147,7 @@ Example for the `subnet_names` variable:
 ```yaml
 subnet_names = {
   ## Required subnets
-  'aks': '<my_aks_subnet_name>', 
+  'aks': '<my_aks_subnet_name>',
   'misc': '<my_misc_subnet_name>',
 
   ## If using ha storage then the following is also required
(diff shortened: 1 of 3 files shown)


docs/CONFIG-VARS.md
```diff
@@ -258,7 +258,8 @@ In addition, you can control the placement for the additional node pools using t
...
-| node_pools_availability_zone | Availability Zone for the additional node pools and the NFS VM, for `storage_type="standard"'| string | "1" | The possible values depend on the region set in the "location" variable. |
+| node_pools_availability_zone | Availability Zone for the additional node pools | string | null | The possible values depend on the region set in the "location" variable.  **Note** Variable superceeded by `node_pools_availability_zones`. |
+| node_pools_availability_zones | Availability Zones for the additional node pools | list of strings | ["1"] | The possible values depend on the region set in the "location" variable. |
 | node_pools_proximity_placement | Co-locates all node pool VMs for improved application performance. | bool | false | Selecting proximity placement imposes an additional constraint on VM creation and can lead to more frequent denials of VM allocation requests. We recommend that you set `node_pools_availability_zone=""` and allocate all required resources at one time by setting `min_nodes` and `max_nodes` to the same value for all node pools.  Additional information: [Proximity Group Placement](./user/ProximityPlacementGroup.md). |
```
riragh, line 263: Need to update the description of variable `node_pools_proximity_placement` as it talks about `node_pools_availability_zone=""`
Also need to update the example file: [sample-input-ppg.tfvars](https://github.com/sassoftware/viya4-iac-azure/blob/main/examples/sample-input-ppg.tfvars#L47)

riragh (dismissed):

docs/CONFIG-VARS.md
riragh: Changes will remain undocumented.

variables.tf
```diff
@@ -359,7 +359,7 @@ variable "netapp_network_features" {
 
 variable "node_pools_availability_zone" {
   type    = string
-  default = null
+  default = "1"
```
jason-matthew, line 362: Not sure this is wanted.  Setting a default conflicts with another portion of this PR

```
 zones = (var.node_pools_availability_zone == "" || var.node_pools_proximity_placement == true) ? [] : (var.node_pools_availability_zone != null) ? [var.node_pools_availability_zone] : var.node_pools_availability_zones`
 ```
 
With this in place, specifying `node_pools_availability_zones` now requires

``` 
node_pools_availability_zone=null
node_pools_availability_zones=["1", "2", "3"]

riragh: As `node_pools_availability_zones` will be an undocumented variable, we are required to keep the default value of the variable `node_pools_availability_zone` same as before. I have adjusted the condition to honor that.

main.tf
```diff
@@ -195,7 +195,7 @@ module "node_pools" {
   max_pods                     = each.value.max_pods == null ? 110 : each.value.max_pods
   node_taints                  = each.value.node_taints
   node_labels                  = each.value.node_labels
-  zones                        = (var.node_pools_availability_zone == "" || var.node_pools_proximity_placement == true) ? [] : [var.node_pools_availability_zone]
+  zones                        = (var.node_pools_availability_zone == "" || var.node_pools_proximity_placement == true) ? [] : (var.node_pools_availability_zone != "1") ? [var.node_pools_availability_zone] : var.node_pools_availability_zones
```
thpang, line 198: Why not just have the logic like this:

```yaml
zones = (var.node_pools_availability_zone == "" || var.node_pools_proximity_placement == true) ? [] :
  length(var.node_pools_availability_zones) != 0 ?
    var.node_pools_availability_zones : [var.node_pools_availability_zone]
```
With this logic we can accurately set the default value of the `node_pools_availability_zones` to `null` and fix the logic when we deprecate the older variable.

Seems like there is a way to clean up this logic unless the value of the first zone in a location is always needed, then the logic can stay.

riragh: Verified by running test with `node_pools_availability_zones` defaulted to `null` and `node_pools_availability_zone` to empty string. The first zone value does not seem to be required. Setting the empty list of string for Zones just sets the `Availability zones` to `None` for the nodes.
I'll update the logic using your suggested condition.

jarpat (approved):

thpang (approved):
LGTM

PR #272, opened 2022-12-09
feat: (IAC-600) Add support for aad rbac

ajeffowens:
Introduces an optional [azure_active_directory_role_based_access_control](https://registry.terraform.io/providers/hashicorp/azurerm/3.26.0/docs/resources/kubernetes_cluster#azure_active_directory_role_based_access_control) block to the azure aks module.  If enabled, the cluster will be integrated with Azure AD.  RBAC permissions within the cluster can then be managed by referencing AAD users and groups.  

Does not change default behavior.

Introduces the following new variables:
- "aks_aad" - this is the primary boolean variable which will control this feature.  None of the following variables will matter if this is false.  
- "aks_aad_tenant_id"
- "aks_aad_admin_group_ids" 
- "aks_aad_azure_rbac_enabled":  disabled by default, undocumented, but here for completeness

More details in internal ticket

### DIFF
--- a/docs/CONFIG-VARS.md
+++ b/docs/CONFIG-VARS.md
@@ -174,6 +174,9 @@ Ubuntu 20.04 LTS is the operating system used on the Jump/NFS servers. Ubuntu cr
 | ssh_public_key | File name of public ssh key for jump and nfs VM | string | "~/.ssh/id_rsa.pub" | Required with `create_jump_vm=true` or `storage_type=standard` |
 | cluster_api_mode | Public or private IP for the cluster api | string | "public" | Valid Values: "public", "private" |
 | aks_cluster_sku_tier | Optimizes api server for cost vs availability | string | "Free" | Valid Values:  "Free", "Paid" | 
+| aks_aad | Whether to integrate this cluster with Azure Active Directory | bool | false | |
+| aks_aad_admin_group_ids | List of Azure Active Directory group ids to be granted admin access over this cluster | list of strings | null | Only relevant when `aks_aad` is true |
+| aks_aad_tenant_id | The Azure tenant id where the aforementioned admin group ids reside.  Defaults to the tenant id of the cluster | string | null | Only relevant when `aks_aad` is true |
 
 ## Node Pools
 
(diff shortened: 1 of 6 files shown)


ajeffowens:
Tests performed (all successful):  

1. deploy w/ aks_aad and aks_aad_admin_group_ids 
2. deploy w/ none of these variables set at all
3. deploy w/ only aks_aad_admin_group_ids set (this should result in same cluster as test 2)

riragh:
Thank you for providing the code for this change. At this time adding support for Azure RBAC is not priority but will be investigated in future. An internal ticket IAC-600 is open for investigation, new PR will be created when priority is set for this feature. Closing this PR.

riragh:
**Additional update:**

`viya4-iac-azure` repo had multiple major updates to support the changes on Azure API, terraform providers and multiple Kubernetes versions. Listing below some of the observations over the course of time which made it difficult to consume the code in this PR.

1. Modules and provider versions were updated, which changed the way how client_key, client_certificate and client_ca_certificates are consumed.
2. For the completeness of the Authentication and Authorization in an AKS cluster, we need additional changes to support the option `Azure AD authentication with Azure RBAC`. However, we are unclear about SAS Viya integration with Azure AD support.
3. One of the major limitation is, starting Kubernetes v1.26 (December/2022), the azure auth provider was removed from the kubectl binary, so user would need an alternative method to be able to authenticate to AKS. The alternative to implement azure authentication is Kubelogin. Kubelogin is a client-go credential plugin that implements Azure AD authentication. It supports credentials plugins to integrate with authentication protocols that are not supported by default by kubectl.

These changes need a lot of code to be added/adjusted for things to work with AAD RBAC.

thpang:
Adding info here for the meeting on 01.31.2024 - Changes here are Azure specific. The IAC/DAC team works very hard to keep things cloud agnostic where possible. Though this PR may work well here, there would need to be some design and discussions around the Active Directory initiative for all the provides being supported. Then we'd need to determine if this change was needed solely for Azure and what the impact would be to our customers both internal and external. We'll discuss more on Wed.

PR #277, opened 2023-01-30
feat: (IAC-804): Use device UUID in mount to avoid NFS issues due to restarts

riragh:
# Changes:
Updated the cloud-init file for nfs to address issue with raid device names being switched and mounts being missing between NFS VM restarts and or terraform (re)apply scenarios

# Tests:
|:----|:----|:----|:----|
|1|NFS VM reboot|Fast 2020|Mounts were correctly bound even after restarts. Viya4 deployment was stable and accessible|
|2|Terraform reapply (see detailed steps in ticket)|Fast 2020|Mounts were correctly bound. Viya4 deployment was stable and accessible|

### DIFF
--- a/files/cloud-init/nfs/cloud-config
+++ b/files/cloud-init/nfs/cloud-config
@@ -41,9 +41,6 @@ runcmd:
   #
   # Create Raid5 Array
   #
-  # NOTE: We need to sleep 5 minutes here to ensure the drives have been attached
-  #
-  # - sleep 300
   - pvcreate $(find /dev/disk/azure/scsi1/ -type l | xargs)
   - vgcreate data-vg01 $(find /dev/disk/azure/scsi1/ -type l | xargs)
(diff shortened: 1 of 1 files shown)


files/cloud-init/nfs/cloud-config
```diff
@@ -41,17 +41,17 @@ runcmd:
   #
   # Create Raid5 Array
   #
-  # NOTE: We need to sleep 5 minutes here to ensure the drives have been attached
-  #
-  # - sleep 300
```
thpang, line 46: So this need just went away? Or does it continue to work given the UUID does not change? Just trying to understand why this was removed

riragh: It has been in the code commented for long. This was removed as cleanup, I can add it back let me know.

jarpat (approved):

thpang (approved):
LGTM

PR #278, opened 2023-02-02
feat: (IAC-638): Updated variable description and validation blocks

riragh:
# Changes:
Added validation for `resource_log_category` and `metric_category` variable. The variables have default so it's not mandatory to provide values but specifying empty list will throw validation errors for both the variables.

# Tests:
Specifying empty lists for `resource_log_category` and `metric_category` throws expected error


jarpat (approved):

thpang (approved):
LGTM

PR #288, opened 2023-02-21
feat: (IAC-886) Enable FIPS support

riragh:
# Changes:
The Federal Information Processing Standard (FIPS) 140-2 is a US government standard that defines minimum security requirements for cryptographic modules in information technology products and systems. This change enables FIPS support for all the nodes and VMs created in the azure cluster.

 - For the cluster to fully support FIPS we need to create all the nodes and VMs with FIPS support enabled. The resource `azurerm_kubernetes_cluster` has flag `fips_enabled` that enables FIPS for the nodes in the cluster's Node Pool. And to update the jump and NFS VMs, this change uses Ubuntu's official FIPS enabled image. See details below.
```
publisher     = "canonical"
offer         = "0001-com-ubuntu-pro-focal-fips"
sku           = "pro-fips-20_04-gen2"
version       = latest
```
 - FIPS support needs Azure CLI version `2.32.0` or later installed and configured, so this change updates the Azure CLI version from `2.24.2` to latest available version `2.45.0`. However, this does not affect the underlying terraform code so this will not be considered as breaking change.

## Additional Changes:
On the currently used kubernetes provider v2.13.0 there is an intermittent timeout issue on service account and secret creation.
This change updates the terraform kubernetes provider version to v2.14.0 which fixes the mentioned issue. See PR [#1833](https://github.com/hashicorp/terraform-provider-kubernetes/pull/1833).

# Tests:
|Scenario|Description|Verification|
|:----|:----|:----|
|1|In terraform.tfvars, set `fips_enabled = true` |All the nodes and VMs (jump and NFS) are created with FIPS enabled|
|2|Defaults, fips_enabled is not set |All the nodes and VMs (jump and NFS) are created without FIPS support enabled|


modules/azurerm_vm/main.tf
```diff
@@ -83,11 +83,20 @@ resource "azurerm_linux_virtual_machine" "vm" {
 
   source_image_reference {
     publisher = var.os_publisher
-    offer     = var.os_offer
-    sku       = var.os_sku
+    offer     = var.fips_enabled ? "0001-com-ubuntu-pro-focal-fips" : var.os_offer
+    sku       = var.fips_enabled ? "pro-fips-20_04-gen2" : var.os_sku
```
thpang, line 87: Does this work each time? Wondering as the Pro versions of Ubuntu usually have some cost or subscription. Guessing SAS may have this but will others? I am guessing they'll need that with their Azure subscription. Just wanting to be sure this isn't working behind the scenes and would break in the field.

modules/azurerm_vm/main.tf
```diff
@@ -83,11 +83,20 @@ resource "azurerm_linux_virtual_machine" "vm" {
...
     version   = var.os_version
   }
 
+  dynamic "plan" {
+    for_each = var.fips_enabled ? [1] : []
+    content {
+      name       = "pro-fips-20_04-gen2"
+      publisher  = "canonical"
+      product    = "0001-com-ubuntu-pro-focal-fips"
```
thpang, line 96: Same comment. This might be in the SAS subscription but others may have to pay for this. Just need to verify.

modules/azurerm_vm/main.tf
riragh: This Ubuntu FIPS image is the first and only FIPS certified image for the public cloud. This is available in Azure Marketplace and should be available for everyone to use, see: https://azuremarketplace.microsoft.com/en-us/marketplace/apps/canonical.0001-com-ubuntu-pro-focal-fips?tab=overview

This does come with some added cost, see pricing here: https://azuremarketplace.microsoft.com/en-us/marketplace/apps/canonical.0001-com-ubuntu-pro-focal-fips?tab=PlansAndPrice

To use this image in your subscription you need to accept terms before deploying cluster. To do so you need run the following az command:
`az vm image terms accept --urn Canonical:0001-com-ubuntu-pro-focal-fips:pro-fips-20_04-gen2:latest --subscription $subscription_id`
See my comment related to this in the code: https://github.com/sassoftware/viya4-iac-azure/blob/IAC-886/variables.tf#L56-L57

modules/azurerm_vm/main.tf
riragh: Same as above.

modules/azurerm_vm/main.tf
riragh: FIPS support itself is not documented yet, this is being introduced as an Experimental feature. Not sure how to proceed with documentation on this specific feature. SAS Viya Platform deployments are not supported on FIPS enabled clusters yet. We will probably be announcing/documenting this feature with same timeline as SAS Viya Platform.

thpang: Got it. Marking as resolved.

thpang (approved):

dhoucgitter (approved):

PR #290, opened 2023-02-24
feat: (IAC-1001) Add ability to set allocation method for public ips of jump and nfs vms

stturc:
My team is looking for the ability to set the DNS name of the jump vm's public IP, so I added the ability to do that. While I was in there, I also added the ability to set the public IP to be Dynamic, so that it will be released and not charged to us while the VM is deallocated.

FYI I have tested this with my changes activated and with no inputs (the defaults).


stturc:
@riragh How can I get this one into the approval flow to be merged in?

riragh:
If you could please provide a little more background/reason for this request then I can create internal ticket for this request and get team approval for the request and the timelines as to when we have this merged.

stturc:
Hey @riragh, the background here is cost savings. We want to be able to have a jumpbox with a public IP for use in specific situations. However since we keep it shutdown most of the time when not in use, the static IP costs us a small fee a month. This adds up when we have multiple clusters. Dynamic IPs are released when the VMs are shutdown, so we save there.

The other portion of this request setting a DNS name for the public IP, is just a nice to have since we won't have static IPs anymore.

For testing, I have tested this as much as I can using the default settings (functionally the same as now), and setting the new variables to make sure they work.

stturc:
FYI @riragh I updated this one again with the latest from staging and main. Hoping it can make the next release!

stturc:
FYI @riragh I updated this one again today. Was there ever an internal ticket created that I can track?

variables.tf
```diff
@@ -322,6 +322,16 @@ variable "create_jump_public_ip" {
   default     = true
 }
 
+variable "jump_public_ip_allocation_method" {
+  default = "Static"
+  type    = string
+}
```
riragh, line 328: Could you please add description for all the new variables introduced in this PR?

riragh (changes requested):

variables.tf
stturc: Done!

stturc:
FYI I can't seem to get these tests to run. @riragh I suspect you need to resolve the changes required message to get them to run.

riragh (approved):

riragh:
> FYI I can't seem to get these tests to run. @riragh I suspect you need to resolve the changes required message to get them to run.

That doesn't seem to be the case. I am looking at what needs to change for the workflows to run.

stturc:
It seems to have run in my repo but is not reporting back.
https://github.com/stturc/viya4-iac-azure/actions/runs/5257876087

riragh (approved):

docs/CONFIG-VARS.md
```diff
@@ -300,6 +302,8 @@ When `storage_type=standard`, a NFS Server VM is created, only when these variab
 | Name | Description | Type | Default | Notes |
 | :--- | ---: | ---: | ---: | ---: |
 | create_nfs_public_ip | Add public ip to the NFS server VM | bool | false | |
+| nfs_public_ip_allocation_method | Defines the allocation method for the public IP address | string | "Static" | Possible values are "Static" or "Dynamic". Only used with `create_nfs_public_ip=true` |
```
thpang, line 305: Looking for a boolean flag here: `create_nfs_static_ip` this matched the other items. What/how do you create a private static ip? Asking as you have public in the name.

modules/azurerm_vm/main.tf
```diff
@@ -7,10 +7,11 @@ resource "azurerm_public_ip" "vm_ip" {
   name                = "${var.name}-public_ip"
   location            = var.azure_rg_location
   resource_group_name = var.azure_rg_name
-  allocation_method   = "Static"
+  allocation_method   = var.public_ip_allocation_method
```
thpang, line 10: code would be cleaner like this
```terraform
create_static_ip = var.create_nfs_static_ip
```
Not quite sure why it needs to be spelled out. Very wordy and extra for a simple flag.

modules/azurerm_vm/main.tf
```diff
@@ -7,10 +7,11 @@ resource "azurerm_public_ip" "vm_ip" {
...
   sku                 = var.vm_zone == null ? "Basic" : "Standard"
   zones               = var.vm_zone == null ? [] : [var.vm_zone]
   tags                = var.tags
+  domain_name_label   = var.public_ip_domain_name_label
```
thpang, line 14: Is the `domain_name_label` always set? Is it optional? Maybe this should be a local. Is this domain only relevant to the VM's being created? Maybe the variable is focused on that. I would drop the `label` element of the variable. It's a domain name for the machine.

thpang (changes requested):
I have made some comments but would like a working session on this one. Lots of extra elements here and adding complexity I don't believe is needed for the end user.

docs/CONFIG-VARS.md
stturc: There is no way to create a private static IP AFAIK, which is why I put public in the name, to show that it was only used with the create_nfs_public_ip flag. But I am happy to follow whatever naming conventions you think are best and show that it only is used in the public IP flow in the documentation.

modules/azurerm_vm/main.tf
stturc: The terraform module needs either Static or Dynamic. So if we created a boolean flag for the end user it would need to be something more like:

allocation_method   = var.create_nfs_static_ip ? "Static" : "Dynamic"

modules/azurerm_vm/main.tf
stturc: I left the label portion of the name since that is what terraform has (I have tried to keep to the terraform conventions as much as possible). Right now this works by either passing in nothing (if it is not set) or an actual value. It could be rewritten to something more like the following for better readability.

domain_name_label   = var.public_ip_domain_name != "" ? var.public_ip_domain_name : ""

docs/CONFIG-VARS.md
thpang: Think we'll be going with `enable_jump_static_ip` and `enable_nfs_static_ip` But that looks good.

thpang (approved):

jarpat (approved):

dhoucgitter (approved):

PR #296, opened 2023-03-06
docs: (IAC-372) Updated the commands related to Service Principal creation

riragh:
## Changes:
Some of the commands in the `AzureHelpTopics.md` document were outdated. Updated the commands related to Service Principal creation.

### DIFF
--- a/docs/user/AzureHelpTopics.md
+++ b/docs/user/AzureHelpTopics.md
@@ -55,7 +55,7 @@ You can create a Service Principal to use with Terraform by taking the following
 az login # follow the instructions given by this command
 
 TF_VAR_client_secret=$(az ad sp create-for-rbac --role "Contributor" --scopes="/subscriptions/$TF_VAR_subscription_id" --name http://$USER --query password --output tsv)
-TF_VAR_client_id=$(az ad sp show --id http://$USER --query appId --output tsv)
+TF_VAR_client_id=$(az ad sp list --display-name http://$USER --query [].appId --output tsv)
 
 echo $TF_VAR_client_id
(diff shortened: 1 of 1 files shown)


thpang (dismissed):

riragh:
For az CLI version `2.37.0` or below, the command `az ad sp credential reset` requires `--name` argument.

While working on FIPS changes I updated the az CLI version on my machine to `2.45.0` which has required argument changed from --name to --id, hence I updated the command to use the --display-name instead. To keep consistency with the az version specified in main branch, I'll revert this command back to original.

dhoucgitter:
> For az CLI version `2.37.0` or below, the command `az ad sp credential reset` requires `--name` argument.
> 
> While working on FIPS changes I updated the az CLI version on my machine to `2.45.0` which has required argument changed from --name to --id, hence I updated the command to use the --display-name instead. To keep consistency with the az version specified in main branch, I'll revert this command back to original.

Just trying to understand, so the commands forms in this PR will work with az cli v2.24.2 mentioned in main README.md, but not az cli 2.45.0?

riragh:
> > For az CLI version `2.37.0` or below, the command `az ad sp credential reset` requires `--name` argument.
> > While working on FIPS changes I updated the az CLI version on my machine to `2.45.0` which has required argument changed from --name to --id, hence I updated the command to use the --display-name instead. To keep consistency with the az version specified in main branch, I'll revert this command back to original.
> 
> Just trying to understand, so the command forms in this PR will work with az cli v2.24.2 mentioned in main README.md, but not az cli 2.45.0?

Yes this command works for all the versions below 2.37.0.

PR #297, opened 2023-03-09
feat: (IAC-950) Add the possibility to specify private dns zone resource id to use for AKS private cluster

morenovj:
Pull request to introduce the necessary code modifications to tackle issue #298


modules/azure_aks/main.tf
```diff
@@ -13,7 +13,7 @@ resource "azurerm_kubernetes_cluster" "aks" {
   kubernetes_version                 = var.kubernetes_version
   api_server_authorized_ip_ranges    = var.aks_cluster_endpoint_public_access_cidrs
   private_cluster_enabled            = var.aks_private_cluster
-  private_dns_zone_id                = var.aks_private_cluster ? "System" : null
+  private_dns_zone_id                = var.aks_private_cluster && var.aks_private_cluster_private_dns_zone_value != "" ? var.aks_private_cluster_private_dns_zone_value : (var.aks_private_cluster ? "System" : null)
```
thpang, line 16: The variable name will need to be changed to `aks_cluster_private_dns_zone_id`

When adjusting for public and private you’ll also need to figure out if you need to use `dns_prefix` for public facing clusters or `dns_prefix_private_cluster` when using private facing clusters. I would imagine this is where the error may be coming from. You’ll need to read through the entire resource document found here: https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster#argument-reference

main.tf
```diff
@@ -141,6 +141,7 @@ module "aks" {
   aks_cluster_node_vm_size                 = var.default_nodepool_vm_type
   aks_cluster_node_admin                   = var.node_vm_admin
   aks_cluster_ssh_public_key               = try(file(var.ssh_public_key), "")
+  aks_private_cluster_private_dns_zone_value = var.aks_private_cluster_private_dns_zone_value
```
thpang, line 144: The variable name will need to be changed to `aks_cluster_private_dns_zone_id`

thpang (changes requested):

morenovj:
Introduced new changes to align with the variable naming. Regarding the `dns_prefix` / `dns_prefix_private_cluster` variables, I added a comparison inside each of them so they are only defined based on the cluster being private or not. I understand that this should have existed before? The code as it is before this PR already let us create private clusters, so I think that `dns_prefix_private_cluster` should have been already defined prior to this PR? Thanks!

modules/azure_aks/main.tf
morenovj: I made a comparison to either define `dns_prefix` or `dns_prefix_private_cluster ` depending on the cluster being private or not. Should have `dns_prefix_private_cluster ` been defined prior to this PR as the code already gave the possibility of creating a private aks cluster? Thank you

morenovj: After trying this whole PR implementation, I think that the previous message was the case as everything seems to be working as expected.

morenovj:
In case it can be useful for the progress of this PR, this implementation was already tested on a use-case to deploy a private AKS cluster with a private DNS zone provided. It appears to be working correctly.

riragh:
## Tests:
|:----|:----|:----|:----|
|1|BYO Private cluster setup with custom Private DNS zone|Fast 2020|Private cluster with custom private dns zone created successfully. Viya4 deployment was successful|
|2|BYO Private cluster setup with AKS managed Private DNS zone: System|Fast 2020|Private cluster with AKS managed System private dns zone created successfully. Viya4 deployment was successful|
|3|Sanity check for non-private (public cluster) with defaults|Fast 2020|Cluster creation and Viya4 deployment were successful|

thpang (approved):

jarpat (approved):

dhoucgitter (approved):

PR #303, opened 2023-03-29
feat: (IAC-712) Support added for PostgreSQL Flexible server with Private access-VNet Integration

riragh:
# Changes:
Azure Database for PostgreSQL - Flexible Server supports two types of mutually exclusive network connectivity methods to connect to your flexible server. The two options are: `Public access (allowed IP addresses)` and `Private access (VNet Integration)`.

This PR adds support for  Private access (VNet Integration). The changes were made following [Microsoft's documentation](https://learn.microsoft.com/en-us/azure/postgresql/flexible-server/concepts-networking#private-access-vnet-integration) and [Terraform guide](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/postgresql_flexible_server)

# Tests:
|:----|:----|:----|:----|
|1|External Postgres with `connectivity_method = "private"`|Fast 2020|PostgreSQL Flexible Server with the private access connectivity method was successfully created. Viya4 deployment stabilized and was accessible|
|2|External Postgres with default `connectivity_method = "public"`|Fast 2020|PostgreSQL Flexible Server with the public access connectivity method was successfully created. Viya4 deployment stabilized and was accessible|


modules/azurerm_postgresql_flex/variables.tf
```diff
@@ -56,10 +56,10 @@ variable "server_version" {
...
 
-variable "public_network_access_enabled" {
-  description = "Whether or not public network access is allowed for this server. Defaults to true"
-  type        = bool
-  default     = true
+variable "connectivity_method" {
+  description = "Network connectivity options to connect to your flexible server. Valid options are 'public' and 'private'. Defaults to public"
+  type        = string
+  default     = "public"
 }
```
thpang, line 63: Was this every surfaced to the end user? If not no comment here. But if it was the change here will be a breaking change.

riragh: No this was never surfaced to end user before, it was an internal variable.
Previously only public access was available so no inputs were required from User. Now with this new change User will have the ability to change the connectivity method, it defaults to public access.

thpang (approved):

jarpat (approved):

PR #308, opened 2023-04-28
feat: (IAC-968) Update Terraform version to latest - 1.4.5

riragh:
# Changes:
This PR updates the default Terraform version to `v1.4.5`

# Tests: WIP


versions.tf
```diff
@@ -3,7 +3,7 @@
 
 terraform {
 
-  required_version = ">= 1.0.0"
+  required_version = ">= 1.4.0"
```
thpang, line 6: Be aware this might be a breaking change for some folks as they run on systems where tf 1.0.0 has been installed

thpang (approved):

versions.tf
dhoucgitter: Expected this to be ">= 1.4.5" since that is the version you tested. Potential risk, even if low to leave it as 1.4.0 since there are 1.4.x versions prior to 1.4.5

riragh: Updated in latest change.

dhoucgitter (approved):

PR #320, opened 2023-06-07
feat: (IAC-1020) Support for Message Broker - Azure Service Bus

riragh:
# Changes:
Support added for Message broker: `Azure service bus` to be used by SAS Viya Platform deployment.
Resources added using:
- https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/servicebus_namespace
- https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/servicebus_namespace_authorization_rule

# Tests:
Verified following on the scenarios listed below, please see internal ticket for details.
- Verified cluster creation was successful.
- Azure Service bus was created. 
- SAS Viya Platform consumed the Azure service bus information from state file correctly. 
- Rabbitmq pods were not created.

**Note:** SAS Viya Platform services are still not ready. Almost all pods stabilized except for few services. Launcher service is failing with `[BROKER_MISSING_USERNAME]` causing other services to fail. This is expected until all the services are updated to use the new Azure Service bus instead of Rabbitmq.

|Scenario|Task|Deploy method|Cadence|
|:----|:----|:----|:----|
|1|Defaults, `create_azure_message_broker= true`, internal Postgres|ansible|fast:2020|
|2|`create_azure_message_broker= true`, external Postgres|ansible|fast:2020|
|3|`create_azure_message_broker= true`, `message_broker_name= "ritika1"`, `aks_cluster_sku_tier = "Standard"`, external Postgres|ansible|fast:2020|

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -252,6 +252,19 @@ module "netapp" {
   depends_on          = [module.vnet]
 }
 
+module "message_broker" {
+  source = "./modules/azurerm_message_broker"
+  count  = var.create_azure_message_broker ? 1 : 0
+
+  resource_group_name     = local.aks_rg.name
+  location                = var.location
+  prefix                  = var.prefix
+  message_broker_sku      = var.message_broker_sku
+  message_broker_name     = var.message_broker_name
+  message_broker_capacity = var.message_broker_capacity
+  tags                    = var.tags
+}
+
 data "external" "git_hash" {
   program = ["files/tools/iac_git_info.sh"]
 }
--- a/modules/azurerm_message_broker/main.tf
+++ b/modules/azurerm_message_broker/main.tf
@@ -1,0 +1,26 @@
+# Copyright © 2020-2023, SAS Institute Inc., Cary, NC, USA. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# Azure Service Bus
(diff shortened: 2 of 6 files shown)


thpang (approved):
Overall I understand this is to replace RabbitMQ and the changes look good. I would have liked to see a more generic approach to the variables and module names along with some investigation as to how these would look across providers. The variable could have been more `message queue` based and generic even though they are utilizing Azure's solution.

riragh:
> Overall I understand this is to replace RabbitMQ and the changes look good. I would have liked to see a more generic approach to the variables and module names along with some investigation as to how these would look across providers. The variable could have been more `message queue` based and generic even though they are utilizing Azure's solution.

After some investigation across cloud for message brokers, these are available options:
```
Azure: Azure Service Bus
AWS: Amazon MQ
GCP: Google Pub/Sub
```
To keep variables and module names generic I went with name as `message broker`. Azure service bus provides option to create queues or topics and SAS Viya Platform creates `Topics`. 
I thought it would be confusing to name the variables/modules as `message queue` when we don't create queues.
I don't mind changing variable/module names, please let me know.

sayeun (approved):
I also did some reading on message broker vs message queue, and AWS, Azure, and GCP.  Similar finding like Ritika reported.

PR #285, opened 2023-02-15
docs: (IAC-910) Update Requirements under the IAC CDS Postgres CONFIG VARS documentation

dhoucgitter:
### Changes
Current document does not specify the requirement to name the CDS Postgres DB 'cds-postgres'. The DAC document has an explanation about the name requirement.

PR #324, opened 2023-06-19
docs: (IAC-1081) Update CDS Postgres URLs in CONFIG VARS documentation

riragh:
## Changes:
The changes required to update the CDS PostgreSQL info was completed in PR #285, Since then the SAS documentation URL to access the details about CDS PostgreSQL has been updated. This ticket is to update that URLs.


docs/CONFIG-VARS.md
```diff
@@ -360,7 +360,7 @@ Each server element, like `foo = {}`, can contain none, some, or all of the para
...
+Multiple SAS offerings require a second PostgreSQL instance referred to as SAS Common Data Store, or CDS PostgreSQL. For more information, see [Common Customizations](https://go.documentation.sas.com/doc/en/itopscdc/default/dplyml0phy0dkr/n08u2yg8tdkb4jn18u8zsi6yfv3d.htm#p0wkxxi9s38zbzn19ukjjaxsc0kl). A list of SAS offerings that require CDS PostgreSQL is provided in [SAS Common Data Store Requirements](https://documentation.sas.com/?cdcId=itopscdc&cdcVersion=default&docsetId=itopssr&docsetTarget=p05lfgkwib3zxbn1t6nyihexp12n.htm#n03wzanutmc6gon1val5fykas9aa). To create and configure an external CDS PostgreSQL instance in addition to the external platform PostgreSQL instance named `default`, specify `cds-postgres` as a second PostgreSQL instance, as shown in the example below.
```
thpang, line 363: Need to know if the new docs URL is going to be a global change? To date all SAS docs have been at go.documentation.sas.com and now it's changing? Need to ask why and if this should be addressed in a future PR to alter or correct what we have. Otherwise things look good.

riragh: I know Susan Pearsall recommended this change but I am not sure why this has changed. @dhoucgitter Any thoughts?

dhoucgitter: @thpang, I believe Susan said that it is actually the same server just a different URL representation from a tool that she uses to generate permanent links. @supear , do I have that right and could you help address Thomas' question regarding if we need to update any of our current uses of go.documentation.sas.com

dhoucgitter: @riragh, in response to your question, the SAS Common Data Store Requirements link was broken which prompted me to ask Susan about it. She generated a new working link for it and then when I noticed that the format looked different from the Common Customizations link she recommended to go ahead and update that one with a similar format.

supear: So, we don't think that you need to remove "go." from SAS Help Center URLs if you have included it. It creeps in there when you copy and paste the URL from the Help Center topic that you want to point to. Help Center URLs are unfortunately kind of confusing. We have a handy conversion tool that you can use; it removes the locale from the URL, which is a good practice. However, if the original URL that you're converting includes a value for "cdcVersion=", then your URL isn't permanent and will eventually become obsolete. The CDC version that needs to be replaced with "default" might also look like this: 
/en/sasadmincdc/v_040/

dhoucgitter: @riragh, you should probably update the Common Customizations link also as I did in https://github.com/sassoftware/viya4-iac-aws/pull/211/files per Susan's comment.

riragh: Updated the other link as well.

dhoucgitter (approved):

sayeun (approved):

thpang (approved):

PR #325, opened 2023-06-21
feat!: (IAC-1021) NFS version default updated to 4.1

riragh:
## Changes:
NFS version default is now updated to v4.1 for both `storage_type = "standard"` and `storage_type = "ha"`.
Previously the `vers` value was hardcoded to `3`, which caused NFS mounts to fail when `netapp_protocols` is set to NFSv4.1. 
This PR reads the value set and replaces the `vers` parameter. Also added note in Troubleshooting.md for file lock issue seen with NFSv3.

### Breaking Change:
This is a breaking change. Upgrading an existing infrastructure that has SAS Viya Platform deployed on it will result in data loss. The changes in this PR destroy and create the NFS VM or Azure NetApp Volume when upgraded, which will result in data loss affecting your existing SAS Viya Platform deployment. 

To avoid SAS Viya Platform deployment data loss perform backup before upgrading, see details in [SAS Documentation](https://go.documentation.sas.com/doc/en/itopscdc/default/calbr/p0cw7yuvwc83znn1igjc16zah2se.htm).

## Tests:

|:----|:----|:----|:----|
|1|Defaults, `storage_type = "standard"`|Fast 2020|NFSv4.1 was used, Jump server was able to mount correctly, SAS Viya Platform deployment stabilized and was accessible|
|1|Defaults, `storage_type = "ha"` and `netapp_protocols = ["NFSv3"]`|Fast 2020|Cluster is created with Azure NetApp using NFS volume v3. Viya4 deployment stabilized and was accessible|
|2|Defaults, `storage_type = "ha"`|Fast 2020|Cluster is created with Azure NetApp using NFS volume v4.1. Viya4 deployment stabilized and was accessible|

### DIFF
--- a/docs/Troubleshooting.md
+++ b/docs/Troubleshooting.md
@@ -1,5 +1,12 @@
 # Troubleshooting
 
+- [Troubleshooting](#troubleshooting)
+  - [Kubernetes Version is not supported in Azure region](#kubernetes-version-is-not-supported-in-azure-region)
+  - [Failure to delete AKS Node Pool](#failure-to-delete-aks-node-pool)
+  - [Import Azure Resource into Terraform state](#import-azure-resource-into-terraform-state)
+  - [Not able to access AKS with kubectl](#not-able-to-access-aks-with-kubectl)
+  - [Azure NetApp Files creation fails](#azure-netapp-files-creation-fails)
+
 ##  Kubernetes Version is not supported in Azure region
 ```bash
 Error: creating Managed Kubernetes Cluster "viya-tst-aks" (Resource Group "viya-tst-rg"): containerservice.ManagedClustersClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: Code="AgentPoolK8sVersionNotSupported" Message="Version 1.18.14 is not supported in this region. Please use [az aks get-versions] command to get the supported version list in this region. For more information, please check https://aka.ms/supported-version-list"
@@ -7,7 +14,7 @@ Error: creating Managed Kubernetes Cluster "viya-tst-aks" (Resource Group "viya-
   on modules/azure_aks/main.tf line 2, in resource "azurerm_kubernetes_cluster" "aks":
    2: resource "azurerm_kubernetes_cluster" "aks" {
 ```
-**Resolution:**
+### Resolution:
 Run this Azure CLI command to get the supported Kubernetes versions in your Azure region and use value for `kubernetes_version` variable in input tfvars.
 ```bash
 az aks get-versions --location <YOUR_AZURE_LOCATION> --output table 
@@ -29,7 +36,7 @@ Error: waiting for the deletion of Node Pool "stateful" (Managed Kubernetes Clus
 Error: A resource with the ID "/subscriptions/REDACTED/resourcegroups/viya-tst-rg/providers/Microsoft.ContainerService/managedClusters/viya-tst-aks/agentPools/stateless" already exists - to be managed via Terraform this resource needs to be imported into the State. Please see the resource documentation for "azurerm_kubernetes_cluster_node_pool" for more information.
 ```
 
-**Resolution:**
+### Resolution:
 
 ```bash
 terraform import -var-file=sample-input.tfvars module.aks.azurerm_kubernetes_cluster.aks '/subscription/REDACTED/../../'
@@ -44,7 +51,7 @@ Error: authorization.RoleAssignmentsClient#Create: Failure responding to request
   18: resource "azurerm_role_assignment" "acr" {
 ```
 
-**Resolution:**
+### Resolution:
 Check values of environment variables - `ARM_* and TF_*`
 
 ## Azure NetApp Files creation fails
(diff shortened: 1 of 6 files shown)


docs/Troubleshooting.md
```diff
@@ -57,5 +65,25 @@ Error: Error creating NetApp Account "sse-vdsdp-ha1-netappaccount" (Resource Gro
...
+
+2. Use Azure NetApp NFS volume version 4.1. This can be done by adding the variable `netapp_protocols` to your terraform.tfvars.
+
+   **Note:** Changing this on existing cluster will result in data loss.
+   
+   Example:
+   ```bash
+   # Storage HA
+   storage_type = "ha"
+   netapp_protocols = ["NFSv4.1"]
```
thpang, line 88: The netap protocol will always be nfs version 4.1 no need to have this as an option or listed here. If we're going to surface this be aware of the support implications. This should be a variable but not listed in the CONFIG-VARS.md file until we're ready to surface the functionality in full for customers and are ready to answer questions.

vms.tf
```diff
@@ -7,9 +7,11 @@ locals {
     : var.storage_type == "ha" ? module.netapp[0].netapp_endpoint : module.nfs[0].private_ip_address
   )
 
+  protocol_version = var.storage_type == "ha" && startswith(var.netapp_protocols[0], "NFS") ? split("v", var.netapp_protocols[0])[1] : "4.1"
```
thpang, line 10: This should always be version 4.1 and have it active. That way the nfs server can support all pervious versions and 4.1 as well. No logic should be needed here.

vms.tf
```diff
@@ -7,9 +7,11 @@ locals {
...
+
   rwx_filestore_path = (var.storage_type == "none"
     ? ""
-    : var.storage_type == "ha" ? module.netapp[0].netapp_path : "/export"
+    : var.storage_type == "ha" ? module.netapp[0].netapp_path : local.protocol_version == "3" ? "/export" : "/"
```
thpang, line 14: Again, looking to keep the directory structure intact here. The exported mount point would and should always be `/export` not `/` This would be very confusing to folks as we now actually export /export but tell them to mount /.

vms.tf
```diff
@@ -19,7 +21,7 @@ locals {
         ["${local.rwx_filestore_endpoint}:${local.rwx_filestore_path}",
           var.jump_rwx_filestore_path,
           "nfs",
-          "_netdev,auto,x-systemd.automount,x-systemd.mount-timeout=10,timeo=14,x-systemd.idle-timeout=1min,relatime,hard,rsize=1048576,wsize=1048576,vers=3,tcp,namlen=255,retrans=2,sec=sys,local_lock=none",
+          "_netdev,auto,x-systemd.automount,x-systemd.mount-timeout=10,timeo=14,x-systemd.idle-timeout=1min,relatime,hard,rsize=1048576,wsize=1048576,nfsvers=${local.protocol_version},tcp,namlen=255,retrans=2,sec=sys,local_lock=none",
```
thpang, line 24: Is the version info needed here? The versioning is controlled on the nfs server side. If we need to force or change the default nfs version, then yes. Otherwise we don't need to include this in the mount command on the external VMs. How will the pods/etc that are using the NFS server going to do this? If this is required we'd need to tell all of the dev teams that the nfsvers value is needed in their nfs mount commands.

thpang (changes requested):
The code here is a quick fix. I am looking for an in depth solution that has little impact on the customer both current and future. We need to understand the full implication of changing to nfs version 4.1 and what/how this would work. I know this one is in draft but again, it simply fixes the issue and does not solve much.

vms.tf
riragh: The exported path is still `/export`, but because it is also the root directory NSFv4 expects you to only mount `/` not `/export`. See the `4. Note with remote NFS paths` in [NFSv4Howto](https://help.ubuntu.com/community/NFSv4Howto#NFSv4_Client)

>They don't work the way they did in NFSv3. NFSv4 has a global root directory and all exported directories are children to it. So what would have been nfs-server:/export/users on NFSv3 is nfs-server:/users on NFSv4, because /export is the root directory.

vms.tf
riragh: `vers` or `nfsvers` info is an optional argument. 
> If no version is specified, NFS uses the highest version supported by the kernel and the mount utility. The client negotiates a suitable version with the server, trying version 4 first, version 3 second, and version 2 last.

In the current arch by removing `vers` results in NFSv3 being used as client is unable to mount `/export` with NFSv4. If we fix the mounting problem with `/export` discussed in the above comment we should be ok to remove version info from here.

vms.tf
riragh: Agreed.

docs/Troubleshooting.md
riragh: I'll fix this.

vms.tf
riragh: After discussion with Thomas, we decided to keep this NFS version check logic in place.

docs/Troubleshooting.md
riragh: NetApp NFS protocol version default is set to v4.1, updated the wording to reflect the same.

thpang (approved):
LGTM - Thanks so very much!

vms.tf
riragh: Additional root directory is added to `/etc/exports`, see updated code in `/files/cloud-init/nfs/cloud-config`.

vms.tf
riragh: Default value for `vers` is set to 4.1. And for NetApp it is determined based on `protocol_version` defaulting to 4.1.

jarpat (approved):

anthoday (approved):

dhoucgitter (approved):

thpang (approved):

PR #327, opened 2023-07-03
fix: (IAC-1090) Fixed jumpvm cloud-init causing Viya deployment failure

riragh:
## Changes:
When a new jump server vm is created, it sets file ownership and permissions for the mounted NFS location which allows Viya services to initialize successfully. Importantly, the file system settings should only be applied once and not repeatedly in the event that the jump vm is destroyed and recreated in the same cluster. This change checks if the `${jump_rwx_filestore_path}/pvs` folder already exists and skips creating the folder and recursively setting ownerships and permissions if it does. If the jump vm is being created for the first time when the pvs folder is absent, creating the folder and setting permissions and ownership will occur an initial time and not thereafter.

This issue originated from AWS, but is applicable for all cloud providers. See the original fix in [AWS PR](https://github.com/sassoftware/viya4-iac-aws/pull/212).

Also, minor update to `outputs.tf` file: updated a missing condition.

## Tests:
Verified the following steps, see details in internal ticket.

#### Steps:
 - Created cluster using IaC Azure
 - Verify infrastructure created successfully
 - Deploy Viya4 using DaC
 - Verify Viya deployment is stabilized
 - Stop Viya. Follow SAS documentation: [Stop a SAS Viya Platform Deployment](https://go.documentation.sas.com/doc/en/itopscdc/v_040/itopssrv/n0pwhguy22yhe0n1d7pgi63mf6pb.htm#p05jsfcz54bsejn1x1r34f0zrag6)
 - Verify all pods are terminated
 - Replace jump server vm using terraform replace command (Alternative to destroy and create), verifyed that the existing jump vm is destroyed and a new one is created.
 - Start Viya following the SAS documentation: [Start a SAS Viya Platform Deployment](https://go.documentation.sas.com/doc/en/itopscdc/v_040/itopssrv/n0pwhguy22yhe0n1d7pgi63mf6pb.htm#p05jsfcz54bsejn1x1r34f0zrag6)
 - Check deployment/pods status verify all Viya pods, services and applications stabilize and are accessible.

PR #330, opened 2023-07-20
8.0.0 - July 20, 2023

riragh:
- #320 
- #325 
- #327 
- #329

### DIFF
--- a/files/cloud-init/jump/cloud-config
+++ b/files/cloud-init/jump/cloud-config
@@ -43,12 +43,15 @@ runcmd:
       # mount the nfs
       #
   -   while [ `df -h | grep "${rwx_filestore_endpoint}:${rwx_filestore_path}" | wc -l` -eq 0 ]; do sleep 5 && mount -a ; done
-      #
-      # Change permissions and owner
-      #
-  -   mkdir -p ${jump_rwx_filestore_path}/pvs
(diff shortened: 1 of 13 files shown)


jarpat (approved):

dhoucgitter (approved):

files/cloud-init/jump/cloud-config
```diff
@@ -43,12 +43,15 @@ runcmd:
...
-  -   $(chown -R nobody:nogroup ${jump_rwx_filestore_path} ; echo)
+  -   if ! [ -d "${jump_rwx_filestore_path}/pvs" ]
+  -   then
+        #
+        # Change permissions and owner
+        #
+  -     mkdir -p ${jump_rwx_filestore_path}/pvs
+  -     $(chmod -fR 777 ${jump_rwx_filestore_path} ; echo)
+  -     $(chown -R nobody:nogroup ${jump_rwx_filestore_path} ; echo)
+  -   fi
```
thpang, line 54: This may be an issue if the customer has set this up prior to running an IAC code and does not have the correct permissions or user assignments. So we might need a doc on this one.

thpang (approved):

files/cloud-init/jump/cloud-config
thpang: Talked with the group and at this time we don't believe this will be an issue.

PR #339, opened 2023-09-25
docs: (IAC-1156) Added Postgres example in separate sample input file

riragh:
# Changes
Previously Postgres example was incorrectly placed in PPG sample input file. This PR creates a separate sample input file for Postgres and removes the incorrectly placed Postgres section from the sample-input-ppg.tfvars file.


thpang (approved):
LGTM

jarpat (approved):

PR #341, opened 2023-10-11
fix: (IAC-1072) Enabling North-South traffic

riragh:
## Changes:
Starting Kubernetes version `1.25` the north-south traffic in Azure stopped working. There were no changes to `azure-ip-masq-agent` or `coredns`. On investigating further found out that changes were made to Azure's standard Loadbalancer. See the following note in [AKS public standard load balancer documentation](https://learn.microsoft.com/en-us/azure/aks/load-balancer-standard#restrict-inbound-traffic-to-specific-ip-ranges)

image

To enable the North-South traffic again for AKS clusters, adding the `aks_pod_cidr` to terraform output for viya4-deployment to consume the changes.

## Tests:
Verified `aks_pod_cidr` was included in terraform.tfstate file

### DIFF
--- a/outputs.tf
+++ b/outputs.tf
@@ -26,6 +26,10 @@ output "aks_cluster_password" {
   sensitive = true
 }
 
+output "aks_pod_cidr" {
+  value = var.aks_pod_cidr
+}
+
 # postgres
 
 output "postgres_servers" {


jarpat (approved):

outputs.tf
```diff
@@ -26,6 +26,10 @@ output "aks_cluster_password" {
   sensitive = true
 }
 
+output "aks_pod_cidr" {
+  value = var.aks_pod_cidr
+}
+
```
thpang, line 32: In the GKE code base we have this - [gke_pod_subnet_cidr](https://github.com/sassoftware/viya4-iac-gcp/blob/main/outputs.tf#L106C9-L106C9) and now we'll have aks_pod_cidr. Would have been better to name the variable `cluster_pod_cidr` to generalize this for each provider. But this will do for now. We'll need another JIRA story around changing and normalizing this.

thpang (approved):

PR #345, opened 2023-10-31
chore: (IAC-1196) Update path pattern to support additional branches

riragh:
## Changes:
Updated the path pattern to support additional branches. Also noticed the child modules were not getting scanned for TFLint so updated the path and running it recursively.
On running TFLint recursively encountered a few errors and warnings which were resolved in this PR. Listing the errors below, changes to resolve the errors were made as per the TFLint recommendations.
~- `Error: Invalid function argument`~
  ~- Invalid value for "path" parameter: no file exists at "/.ssh/id_rsa.pub". We do not distribute this file as part of configuration code so setting the default path value returns error. Change was made similar to what we have in GCP. the files `locals.tf`, `variables.tf` and `CONFIG-VARS.md` were updated to resolve this error.~
- `terraform_unused_declarations`
  - The unused variables `aks_cluster_rg_id`, `create_netapp`, `vnet_name`, `private_dns_zone_id` were removed.
- `terraform_deprecated_index`
  - We were using the legacy form of splat expressions, which was updated to resolve the warning.

See the last commit on this PR for the three checks running successfully. 

## Tests
| Scenario | Provider | K8s Version             | Order  | Cadence   | Orchestration |
|----------|----------|-------------------------|--------|-----------|---------------|
| 1        | Azure | 1.26  | ****** | fast:2020 | DO            |


jarpat (approved):

docs/CONFIG-VARS.md
```diff
@@ -191,7 +191,7 @@ Ubuntu 20.04 LTS is the operating system used on the Jump/NFS servers. Ubuntu cr
 | jump_rwx_filestore_path | File store mount point on jump server | string | "/viya-share" | This location cannot include `/mnt` as its root location. This disk is ephemeral on Ubuntu, which is the operating system being used for the jump/NFS servers. |
 | tags | Map of common tags to be placed on all Azure resources created by this script | map | { project_name = "sasviya4", environment = "dev" } | |
 | aks_identity | Use UserAssignedIdentity or Service Principal as  [AKS identity](https://docs.microsoft.com/en-us/azure/aks/concepts-identity) | string | "uai" | A value of `uai` wil create a Managed Identity based on the permissions of the authenticated user or use [`AKS_UAI_NAME`](#use-existing), if set. A value of `sp` will use values from [`CLIENT_ID`/`CLIENT_SECRET`](#azure-authentication), if set. |
-| ssh_public_key | File name of public ssh key for jump and nfs VM | string | "~/.ssh/id_rsa.pub" | Required with `create_jump_vm=true` or `storage_type=standard` |
+| ssh_public_key | File name of public ssh key for jump and nfs VM | string | null | Required with `create_jump_vm=true` or `storage_type=standard` |
```
thpang, line 194: This can now be a breaking change as all our code looks for this and the users might not be used to added or adjusting this line. What/how are we replacing this default?

main.tf
```diff
@@ -131,7 +131,6 @@ module "aks" {
 
   aks_cluster_name                         = "${var.prefix}-aks"
   aks_cluster_rg                           = local.aks_rg.name
-  aks_cluster_rg_id                        = local.aks_rg.id
```
thpang, line 134: No longer needed? This is unclear.

main.tf
```diff
@@ -241,7 +240,6 @@ module "netapp" {
   prefix              = var.prefix
   resource_group_name = local.aks_rg.name
   location            = var.location
-  vnet_name           = module.vnet.name
```
thpang, line 244: No longer needed? This is unclear.

modules/azure_aks/variables.tf
```diff
@@ -11,11 +11,6 @@ variable "aks_cluster_rg" {
   type        = string
 }
 
-variable "aks_cluster_rg_id" {
-  description = "The `azurerm_kubernetes_cluster`'s id."
-  type        = string
-}
-
```
thpang, line 18: No longer needed? This is unclear.

modules/azurerm_netapp/variables.tf
```diff
@@ -1,12 +1,6 @@
 # Copyright © 2020-2023, SAS Institute Inc., Cary, NC, USA. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-variable create_netapp {
-  description = "Boolean flag to create Azure NetApp Files"
-  type        = bool
-  default     = false
-}
-
 variable "prefix" {
```
thpang, line 4: No longer needed? This is unclear.

modules/azurerm_netapp/variables.tf
```diff
@@ -22,11 +16,6 @@ variable "location" {
   type        = string
 }
 
-variable "vnet_name" {
-  description = "Azure Virtual Network"
-  type        = string
-}
-
```
thpang, line 29: No longer needed? This is unclear.

modules/azurerm_postgresql_flex/variables.tf
```diff
@@ -98,9 +98,3 @@ variable "delegated_subnet_id" {
   description = "The ID of the virtual network subnet to create the PostgreSQL Flexible Server. The provided subnet should not have any other resource deployed in it and this subnet will be delegated to the PostgreSQL Flexible Server, if not already delegated. Changing this forces a new PostgreSQL Flexible Server to be created."
   type        = string
 }
-
-variable "private_dns_zone_id" {
-  description = "The ID of the private DNS zone to create the PostgreSQL Flexible Server. Changing this forces a new PostgreSQL Flexible Server to be created."
-  type        = string
-  default     = null
-}
```
thpang, line 106: No longer needed? This is unclear.

variables.tf
```diff
@@ -79,7 +79,7 @@ variable "fips_enabled" {
 variable "ssh_public_key" {
   description = "A custom ssh key to control access to the AKS cluster. Changing this forces a new resource to be created."
   type        = string
-  default     = "~/.ssh/id_rsa.pub"
+  default     = null
```
thpang, line 82: Need new appropriate default value or local that would mirror what we had in the past.

thpang (changes requested):

main.tf
thpang: Resolved from issue description and linting.

modules/azure_aks/variables.tf
thpang: Resolved from issue description and linting.

modules/azurerm_netapp/variables.tf
thpang: Resolved from issue description and linting.

modules/azurerm_netapp/variables.tf
riragh: Variable was specified in the `variables.tf` file but was not used in `main.tf`, hence tflint throws `terraform_unused_declarations` warning. Removing this does not affect the functionality.

modules/azurerm_postgresql_flex/variables.tf
riragh: same as above.
Variable was specified in the `variables.tf` file but was not used in main.tf, hence tflint throws `terraform_unused_declarations` warning. Removing this does not affect the functionality.

main.tf
riragh: The value for `vnet_name` set here is not used in the `azurerm_netapp/main.tf`.  Removing this did not affect the functionality.

docs/CONFIG-VARS.md
riragh: TFLint throws error when setting the value to a non-existent file in `locals.tf`. I adjusted the logic in [locals.tf ](https://github.com/sassoftware/viya4-iac-azure/blob/IAC-1196/locals.tf#L7) similar to how this is set in [GCP](https://github.com/sassoftware/viya4-iac-gcp/blob/main/locals.tf#L23). GCP also sets the default value for `ssh_public_key` as null. This change wasn't synced across cloud. 

Following is the error I received from TFLint:
```
Error: Invalid function argument

  on locals.tf line 8, in locals:
   8:     ? file(var.ssh_public_key)

with var.ssh_public_key as "~/.ssh/id_rsa.pub".

Invalid value for "path" parameter: no file exists at "~/.ssh/id_rsa.pub"; this function works only with files that are distributed as part of the configuration source code, so if this file will be created by a resource in this configuration you must instead obtain this result from an attribute of that resource.

Error: Process completed with exit code 1.
```
We have always stated this as required when `create_jump_vm=true` or `storage_type=standard`. Also all our sample files  have this variable set. But you might be right because there was a default in place before folks might have overlooked at this requirement.
This one is tricky because we do not generate keypair and rather depend on user to provide it.
Will reach out to you to discuss more about this.

thpang: I think over the short-term we should leave this and have the exception setup in the .tflint.hcl file to over-ride this error. The information is correct, but we'd need to come up with a more elegant solution in fixing this issue given its past usage. Or simply change this when we're going to have our next major release with other breaking changes.

riragh: The changes for variable `ssh_public_key` were backed out. TFLint does not seem to be throwing the error previously received. Pushing up for re-review.

variables.tf
riragh: The changes for variable `ssh_public_key` were backed out. TFLint does not seem to be throwing the error previously received. Pushing up for re-review.

thpang (approved):
LGTM

PR #352, opened 2023-11-22
feat: (IAC-1221) Update terraform and terraform provider versions

riragh:
## Changes:
To remediate high security vulnerabilities updated Terraform version, few terraform provider versions and Azure CLI version to latest. No breaking changes.

#### Version updates:
* Terraform 1.4.5 -> 1.6.4
* Azure CLI 2.48.1 -> 2.54.0
* Kubectl 1.26.6 -> 1.26.10

#### Provider version updates:
* azurerm 3.64.0 -> 3.81.0
* azuread 2.39.0 -> 2.46.0
* external 2.3.1 -> 2.3.2
* null 3.2.1 -> 3.2.2
* kubernetes 2.20.0 -> 2.23.0

## Tests: WIP
Verified following scenarios, see internal ticket for details


jarpat (approved):

Dockerfile
```diff
@@ -12,7 +12,7 @@ COPY . .
 
 RUN apk update \
   && apk upgrade \
-  && apk add --no-cache git openssh \
+  && apk add --no-cache git openssh curl\
```
thpang, line 15: What changed here? Wondering why we now need curl included and we did not before and the code worked below on line 16.

riragh: Docker build command failed with error: `curl: not found`. 
Something changed on the latest alpine image, curl is not pre-installed. I didn't find any note on why or how it was changed, but explicitly adding curl resolves the error.

thpang: Looks good. Newer version of Azure CLI seems to be locking things down. Thx for the update.

thpang (approved):

PR #372, opened 2024-03-22
feat: (IAC-1386) EncryptAtHost changes for NIST

riragh:
## Changes:
This PR adds EncryptAtHost requirements in the VMs and AKS nodes.

- Virtual machines and virtual machine scale sets should have encryption at host enabled.
- Temp disks and cache for agent node pools in AKS clusters should be encrypted at host.

This is an optional features, to enable encryption at host for nodes and VMs you would set the following variables:

| Name | Description | Type | Default |
| :--- | :--- | ---: | ---: |
| enable_vm_host_encryption | Setting this variable enables all of the disks (including the temp disk) attached to this Virtual Machine be encrypted by enabling Encryption at Host. This setting applies to both Jump and NFS VM. | bool | false |
| vm_disk_encryption_set_id | The ID of the Disk Encryption Set which should be used to Encrypt this OS Disk. This setting applies to both Jump and NFS VM. | string | null |
| aks_cluster_enable_host_encryption | Enables host encryption on all the nodes in the Node Pool. | bool | false |
| aks_node_disk_encryption_set_id | The ID of the Disk Encryption Set which should be used for the Nodes and Volumes. Changing this forces a new resource to be created. | string | null |

By default, when using AKS, OS and data disks use server-side encryption with platform-managed keys. The caches for these disks are encrypted at rest with platform-managed keys. You can specify your own managed keys following [Bring your own keys (BYOK) with Azure disks in Azure Kubernetes Service](https://learn.microsoft.com/en-us/azure/aks/azure-disk-customer-managed-keys). The caches for these disks are also encrypted using the key you specify.

When using customer-managed-keys, ensure you have the proper AKS credentials. The managed identity needs to have contributor access to the resource group where the diskencryptionset is deployed. Otherwise, you'll get an error suggesting that the managed identity does not have permissions.
Also, make sure the VNet of your deployment has access to the Key Vault which has the diskencryptionset. 
Note: Adding the subnet service endpoints makes the process smooth.

## Tests:
|:----|:----|:----|:----|
|1|Defaults, no changes|Fast 2020|Transparent to user, no changes|
|2|Set `enable_vm_host_encryption = true`, `aks_cluster_enable_host_encryption = true`|fast:2020| Encryption at host enabled with platform-managed keys on VM and nodes in nodepools |
|3|Set `enable_vm_host_encryption = true`, `vm_disk_encryption_set_id = "<disk_encryption_set_id>"`, `aks_cluster_enable_host_encryption = true`, `aks_node_disk_encryption_set_id = "<disk_encryption_set_id>"`|fast:2020| Encryption at host enabled with customer-managed keys on VM and nodes in nodepools |

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -144,7 +144,8 @@ module "aks" {
   aks_cluster_max_pods                     = var.default_nodepool_max_pods
   aks_cluster_os_disk_size                 = var.default_nodepool_os_disk_size
   aks_cluster_node_vm_size                 = var.default_nodepool_vm_type
-  aks_cluster_enable_host_encryption       = var.enable_default_nodepool_host_encryption
+  aks_cluster_enable_host_encryption       = var.aks_cluster_enable_host_encryption
+  aks_node_disk_encryption_set_id          = var.aks_node_disk_encryption_set_id
   aks_cluster_node_admin                   = var.node_vm_admin
   aks_cluster_ssh_public_key               = try(file(var.ssh_public_key), "")
   aks_cluster_private_dns_zone_id          = var.aks_cluster_private_dns_zone_id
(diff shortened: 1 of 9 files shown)


thpang (commented):
Are these breaking changes? I see where we are adding items into the node pool definitions and how the vms are being created as well.

riragh:
> Are these breaking changes? I see where we are adding items into the node pool definitions and how the vms are being created as well.

No not breaking change. The default values are set for the variables as `false` and `null`, which guards this. So it doesn't affect the existing user.

jarpat (approved):

thpang (approved):
LGTM

PR #395, opened 2024-08-26
set 'sku' to 'Standard'

jasonthemartyr:
Default NSG settings for Azure VM's for "basic" SKU is open to all. This PR explicitly sets the default to "Standard"


riragh:
Sorry for the delayed response, Thank you for providing with PR. The sku for Azure VM is now set to 'standard' by default. This change is present in main branch. Update was made in the bigger effort of major version update for Terraform azurerm provider version from v3.92.0 to v4.3.0.
Closing this PR.

PR #412, opened 2024-11-18
fix: permissions set in dockerfile (PSKD-903)

illianov:


illianov:
Closing this PR since I branched from main instead of staging.

PR #417, opened 2024-12-27
chore(deps): update default vm generations off v3

### LINKED ISSUE #404: bump suggested instance generations to v5
I see the change to not use the machine with the extra disk, but can we leave the default generation as v5 for the newer chipset and better I/O?

> Why change this default value to a v3 instance from a v5? I see you switch from a D4ds to a D4s, but using a D4s_v5 gives you much more I/O for the same price, and its a newer machine generation. 

 _Originally posted by @Carus11 in [651c565](https://github.com/sassoftware/viya4-iac-azure/commit/651c565241d52bb9fb5a9b25a9168b3f0f102d1b#r148102584)_

Carus11:
Fixes #404

The v3 instances are now hidden away in the Microsoft docs under previous and retired generations. The default generation produced by sizing these days are the v5's.

### DIFF
--- a/docs/CONFIG-VARS.md
+++ b/docs/CONFIG-VARS.md
@@ -368,7 +368,7 @@ Each server element, like `foo = {}`, can contain none, some, or all of the para
 
 | Name | Description | Type | Default | Notes |
 | :--- | ---: | ---: | ---: | ---: |
-| sku_name| The SKU Name for the PostgreSQL Flexible Server | string | "GP_Standard_D4ds_v5" | The name pattern is the SKU, followed by the tier + family + cores (e.g. B_Standard_B1ms, GP_Standard_D2s_v3, MO_Standard_E4s_v3).|
+| sku_name| The SKU Name for the PostgreSQL Flexible Server | string | "GP_Standard_D4ds_v5" | The name pattern is the SKU, followed by the tier + family + cores (e.g. B_Standard_B1ms, GP_Standard_D2s_v5, MO_Standard_E4s_v5).|
 | storage_mb | The max storage allowed for the PostgreSQL Flexible Server | number | 131072 | Possible values are 32768, 65536, 131072, 262144, 524288, 1048576, 2097152, 4194304, 8388608, 16777216, and 33554432. |
 | backup_retention_days | Backup retention days for the PostgreSQL Flexible server | number | 7 | Supported values are between 7 and 35 days. |
 | geo_redundant_backup_enabled | Enable Geo-redundant or not for server backup | bool | false | Not supported for the basic tier. |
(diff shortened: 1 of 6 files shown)


dhoucgitter:
Hi @Carus11 , expecting that this request would have us update any surviving _v3 references in our sample.*.tfvars files under the examples folder to _v5 types also?

Carus11:
Good catch, updated those in the branch. 
Additionally, I updated the connect sample even though the documentation states:
> Note: The functionality associated with the optional Connect workload class has been deprecated and is no longer available in 2024.09 and later.

Thats probably worth of another task to clean it up.

PhanikumarThodupunoori:
Hi Carus,

Given that the _v5 SKU is not available in many regions, I suggest update the PR to use the _v4 SKU, which is available in most of the regions, to avoid deployment issues. Additionally, the current default values in variables.tf are expected to work across all regions. Setting _v5 SKU as the default conflicts with this expectation.

Please provide your feedback, and let us know if you can update the PR accordingly. Once updated, Please let us know. Thank you!

Carus11:
Just running through and checking from the product availability graphs,
- Dsv5 has greater availability worldwide than Dsv4
- Esv5 has greater availability worldwide than Esv4

The v5 instances were significantly upgraded in their ability to work with storage over the network. The are a minimum generation version I suggest to customers because of this. The I/O improvements are soo great that I would prefer to leave v5 as the example generation to strive for when available. Its perfectly reasonable for users to adjust the instances to suit their specific constraints when working in one of the few regions with constrained product lists.

Carus11:
>  Additionally, the current default values in variables.tf are expected to work across all regions

If this is a requirement, then it is expected to work across all regions then leaving them on v3 would be the closest method of satisfying that, as its only missing it appears in one region in china, and the one new zealand region which doesnt have many other products listed either.

Carus11:
V3 is the original value I was hoping to move forward from. But due to the requirement mentioned of needing the greatest global availability as the default I will simply close this PR.

PhanikumarThodupunoori:
We’ve reviewed this further and concluded that proceeding with the v5 changes for the servers is the right direction to align with our objectives. However, for flexible SKUs, we will keep them as v3 for better availability.
 
We’ll make a small commit to incorporate these updates before merging.

PhanikumarThodupunoori:
We made the change to the flex server instance size to revert it back to v3. We've been able to create clusters using the sample-input-minimal and sample-input-connect files and they successfully use the v5 instance size.

saschjmil (approved):
LGTM. Thanks for the update Carus!

PhanikumarThodupunoori (approved):

PR #424, opened 2025-01-10
PMFR4PLTFND-1332: netapp_enable_cmk_encryption

ajeffowens:
This PR introduces three new variables which should enable CMK encryption on the ANF account.  

```
variable "netapp_enable_cmk_encryption" {
  description = "Setting this variable to true enables CMK encryption on the netapp account.  Only relevant when storage_type=ha."
  type        = bool
  default     = false
}

variable "netapp_cmk_encryption_key_id" {
  description = "The ID of the key in keyvault to Encrypt ANF with (i.e. https://<keyvault-name>.vault.azure.net/keys/<key-name>).  Must exist before running terraform.  Only relevant when storage_type=ha.  Required if enable_anf_cmk_encryption is true."
  type        = string
  default     = null
}

variable "netapp_cmk_encryption_key_uai" {
  description = "The user assigned identity that will be used to access the key (i.e. /subscriptions/<sub>/resourceGroups/<rg>/providers/Microsoft.ManagedIdentity/userAssignedIdentities/<uai name>).  Must exist and have Key Vault Crypto Service Encryption User permission on the keyvault before running terraform.  Only relevant when storage_type=ha.  Required if enable_anf_cmk_encryption is true."
  type        = string
  default     = null
}
```

The vars are not doc-ed to follow the pattern of other encryption-related vars.  

This should be a non-breaking change.  If netapp_enable_cmk_encryption is false, and/or if the variables are omitted entirely, the ANF module should continue to work exactly as before.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -259,7 +259,7 @@ module "netapp" {
   resource_group_name = local.aks_rg.name
   location            = var.location
   subnet_id           = module.vnet.subnets["netapp"].id
-  network_features    = var.netapp_network_features
+  network_features    = var.netapp_enable_cmk_encryption ? "Standard" : var.netapp_network_features
   service_level       = var.netapp_service_level
   size_in_tb          = var.netapp_size_in_tb
   protocols           = var.netapp_protocols
@@ -269,6 +269,10 @@ module "netapp" {
   depends_on          = [module.vnet]
 
   community_netapp_volume_size = var.community_netapp_volume_size
+
+  netapp_enable_cmk_encryption  = var.netapp_enable_cmk_encryption
+  netapp_cmk_encryption_key_id  = var.netapp_cmk_encryption_key_id
+  netapp_cmk_encryption_key_uai = var.netapp_cmk_encryption_key_uai
 }
 
 data "external" "git_hash" {
--- a/modules/azurerm_netapp/main.tf
+++ b/modules/azurerm_netapp/main.tf
@@ -10,6 +10,22 @@ resource "azurerm_netapp_account" "anf" {
   location            = var.location
   resource_group_name = var.resource_group_name
   tags                = var.tags
+
+  dynamic "identity" {
+    for_each = var.netapp_enable_cmk_encryption ? [1] : []
+    content {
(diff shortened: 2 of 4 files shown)


Carus11:
Hey Jeff, you will need to amend your commit and use --signoff if using the git cli to adhere to the DCO requirements https://github.com/sassoftware/viya4-iac-azure/pull/424/checks?check_run_id=35447478424

dhoucgitter:
Hi @ajeffowens, I'm looking into this PR. Did you follow a set of published steps to create the key and set up permissions for the uai to access the keyvault when you tested this?

dhoucgitter:
Hi @ajeffowens, are you able to share the terraform-input.tfvars file that you used for your testing? AS part of the steps to create the key vault, it looks like this will need to be a BYON scenario where you create a pre-existing VNet to both create your AKS cluster in and that your key vault private endpoint can live in, does that sound similar to the configuration that you used?

ajeffowens:
You are correct, David, that this implies not only a BYOKeyvault/key, but also BYON/pep/mi.  This code would not support assigning the required permissions to the MI, let alone creating it (let alone creating the KV).  I do not know whether this would work with a public keyvault, but that is probably an unlikely scenario to support.  

In our case, we always pre-create the MI, then the keyvault and keys, then the Vnet and PE.  So we use standard BYON inputs to iac + the three additional inputs defined by this PR.  

Similar pattern to the `aks_cluster_enable_host_encryption`, `enable_vm_host_encryption`, `vm_disk_encryption_set_id`, and `aks_node_disk_encryption_set_id` vars.  (where keyvault/key/des/mi/rbac must be sorted out ahead of time).

saschjmil:
Hey @ajeffowens, sorry for closing and re-opening your PR. I deleted the staging branch, which closed all PR's targeting it. I've updated your PR to point to the main branch.

ajeffowens:
Cool, and no worries.  I will get my branch rebased to main as well

ancgup0628:
Hi @ajeffowens , could you please confirm if this is still required or we can close this request.

PR #432, opened 2025-02-18
test: refactor tests using table tests (PSKD-1363)

illianov:
This PR is a rough draft in the effort to refactor the terraform tests to be more modular.


ephesused:
I suspect this PR should be opened against `pr-pskd-1145`, not `main`.

test/default_integration_plan_new_test.go
```diff
@@ -0,0 +1,104 @@
...
+	"testing"
+
+	"github.com/gruntwork-io/terratest/modules/random"
+	"github.com/gruntwork-io/terratest/modules/terraform"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"test/validation"
+)
+
+type TestParams struct {
```
ephesused, line 16: I think this will be cleaner if you move the struct definition and the test definitions into TestPlanDefaults. Among other reasons, the test definition move will enable per-test invocations in IDEs that look for the common golang subtest pattern.

ephesused (changes requested):
I really like how this looks, thanks!

test/default_integration_plan_new_test.go
ephesused: I suspect you can make these struct members private, too.

illianov:
This PR can be closed. The table test harness commits have been applied to the pr-pskd-1145 branch. This pr-pskd-1363 branch is no longer needed.

PR #436, opened 2025-02-19
test: rbac tests using jsonpath (PSKD-1378)

kylhia:


saschjmil (approved):
Looks good to me. Thanks for making it easier to sort through these turtles.

iadomi (approved):

iadomi:
@saschjmil the github workflow for some reason

saschjmil:
It looks like the `go test` hit a runtime error due to an undefined function. I'm not sure why since I see the function defined in the helpers.go file. Maybe because helpers.go is missing to //go:build tag at the top? However, the output files were still generated and the log parser worked as expected. 

`Test Results:`
`Suite: command-line-arguments | Failures: 1  `
`exit status 1`

I think this demonstrates that our workflow is doing what it's supposed to (although we could probably modify the reporting to make the failures more obvious).

PR #439, opened 2025-02-21
test: table test framework (PSKD-1363)

illianov:


.github/workflows/default_plan_unit_tests.yml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Set ssh keys for github remote download
+        working-directory: test
+        run: |
+          mkdir ~/.ssh
+          echo "${{ secrets.SSH_GITHUB }}" > ~/.ssh/id_rsa.pub
+          chmod 600 ~/.ssh/id_rsa.pub
+      - name: Run Tests
+        working-directory: test
+        run: |
+          go test ./... -timeout 30m | tee test_output.log
```
ephesused, line 36: Small suggestion: By default, I like tests that are quiet. For automated runs like github workflows, though, I'd prefer to have `-v` on the `go test` call so that the logs are clear about which tests actually ran. The verbosity can help reviewers ensure things went as planned in the workflow (instead of leaving open the possibility that something quietly was skipped). It also can be helpful if things are running slowly, and there's a question about whether they're moving but slow versus hung.

saschjmil: Thanks for the suggestion. @iadomi and I have been updating the workflow in the [pr-pskd-1358](https://github.com/sassoftware/viya4-iac-azure/tree/pr-pskd-1358) branch, but I will update it to run in verbose mode.

joboon (approved):

dhoucgitter (approved):

PR #440, opened 2025-02-25
feat: Refactor ACR tests (PSKD-1400)

ndickens:
I wanted to add some functionality to the test cases to:
 - be able to specify the presence or absence of state resources
 - perform element matching for arrays 
 - display an error message
 - in general have more control over how information is retrieved from the plan and verified

To that end I refactored TestCase to be an interface.  A test could run a collection of test cases that would all implement RunTest and the underlying test type would inform how data is retrieved and verified.  If there isn't a test type that suits your need then just create a new one and implement its RunTest.  getExpectedFromPlan() can also be modified and extended to be able to look in different sections of the plan struct as needed.

Feedback is welcome.


test/acr_test.go
```diff
@@ -1,85 +1,142 @@
...
-	// Validate geo-replication locations
-	actualGeoReplications, err := getJsonPathFromStateResource(acrResource, "{$.georeplications[*].location}")
-	assert.NoError(t, err)
-	expectedGeoReplications := variables["container_registry_geo_replica_locs"].([]string)
-	assert.ElementsMatch(t, expectedGeoReplications, strings.Fields(actualGeoReplications), "Geo-replications do not match expected values")
+func ACRNameContains(name string) TestCase {
+	return &StringContainsTestCase{
+		expected: name,
+		path:     []string{"azurerm_container_registry.acr[0]", "{$.name}"},
+		message:  "ACR name does not contain 'acr'",
```
ephesused, line 108: Nit: Instead of 'acr', this message should use the value of `name`.

test/test_case.go
```diff
@@ -0,0 +1,104 @@
...
+func (testCase *ElementsMatchTestCase) RunTest(t *testing.T, plan *terraform.PlanStruct) {
+	actual, err := getExpectedFromPlan(plan, testCase.path, "string_array")
+	require.NoError(t, err)
+	assert.ElementsMatch(t, actual, testCase.expected, testCase.message)
+}
+
+func getExpectedFromPlan(plan *terraform.PlanStruct, path []string, expectedType string) (any, error) {
+	valuesMap := plan.ResourcePlannedValuesMap[path[0]]
+
+	if expectedType == "string" {
```
ephesused, line 86: Non-blocking: Maybe swap to a `switch` and use an enum for these types.

PR #451, opened 2025-03-14
feat: experimental spot node pools

### LINKED ISSUE #399: Spot node pools
### Is your feature request related to a problem? Please describe.

Not related to a problem

### Describe the solution you'd like

The spot node pool functionality is partially implemented yet commented out. Enabling the use of spot nodes would greatly improve the cost situation when running these environments in short term deployments for technical validation, or demos.

This could be added also with a very clear experimental label

### Describe alternatives you've considered

Considered using the small sizings. But there is a floor to the size of a viya environment which is still fairly costly.

### Additional context

I understand that there has been some concern in the past about this feature not being a supported pattern, but I believe the option should be available for those of us creating environments which we know will be unstable because of how spot instances work. The configuration variable could use a notice regarding the support status to drive this home.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Carus11:
closes #399 

This PR introduces some configuration options to allow the experimental spot node pools. This is useful when running in short term lab environments and testing minor use cases. 

I imagine it should include some additional warnings around the lack of support of this configuration. Im open to any feedback on how to best communicate that.


saschjmil:
Hey @Carus11, thanks for opening this. I've created an internal ticket to verify the work. I'll reach out to you to learn more about the proposed change and brainstorm ways we can document it.

Carus11:
Would it be more palatable if the config variable names were prefixed with `EXPERIMENTAL_` in order to delineate them from a supported configuration?

Carus11:
WIP

Carus11:
The analysis check fails because of the recently released bad version of hashicorp/local, there is an issue here:
https://github.com/hashicorp/terraform-provider-local/issues/408

docs/community/community_config_vars.md
```diff
@@ -20,4 +20,7 @@ Here is a table with the variables you would use to configure them
 
 | Name | Description | Type | Default | Release Added | Notes |
 | :--- | ---: | ---: | ---: | ---: | ---: |
-| contrib_enable_spot_nodes | Enable spot nodes | bool | false | 10.3.0 | |
+| community_priority | (Optional) The Priority for Virtual Machines within the Virtual Machine Scale Set that powers this Node Pool. Possible values are Regular and Spot. Defaults to Regular. Changing this forces a new resource to be created. | string | `Regular` | 10.3.0 | Changing this to Spot enables the Spot node pool functionality |
+| community_eviction_policy | (Optional) The Eviction Policy which should be used for Virtual Machines within the Virtual Machine Scale Set powering this Node Pool. Possible values are Deallocate and Delete. Changing this forces a new resource to be created. | string | `Delete` | 10.3.0 | |
+| community_spot_max_price | (Optional) The maximum price you're willing to pay in USD per Virtual Machine. Valid values are -1 (the current on-demand price for a Virtual Machine) or a positive value with up to five decimal places. Changing this forces a new resource to be created. | string | `-1` | |
```
saschjmil, line 25: Adding "Release Added" column.
```suggestion
| community_spot_max_price | (Optional) The maximum price you're willing to pay in USD per Virtual Machine. Valid values are -1 (the current on-demand price for a Virtual Machine) or a positive value with up to five decimal places. Changing this forces a new resource to be created. | string | `-1` | 10.3.0 | |
```

docs/community/community_config_vars.md
```diff
@@ -20,4 +20,7 @@ Here is a table with the variables you would use to configure them
```
saschjmil, line 20: Could you replace the "Here is some info about spot nodes" and "here is a warning about why they might cause issues" lines with some more information (line 15 and 17)? It doesn't need to be too detailed. Perhaps something like "Spot Nodes are ___. They can be beneficial because ___. They can cause instability because ___. More info about them can be found here: <link to Azure doc>"

variables.tf
```diff
@@ -568,11 +568,15 @@ variable "node_pools" {
     max_pods     = string
     node_taints  = list(string)
     node_labels  = map(string)
+    community_priority     = optional(string, "Regular")
+    community_eviction_policy = optional(string)
+    community_spot_max_price = optional(string)
```
saschjmil, line 573: Just wanting to double check how this compares to current functionality:
* Regular priority is how it currently functions. 
* An empty eviction_policy will use "Delete", which is how it currently functions.
* An empty spot_max_price will default to "-1", which means no max price. This is how it currently functions.

Nothing in IAC changes is these three variables aren't set. Is that all correct?

modules/aks_node_pool/variables.tf
```diff
@@ -116,6 +116,22 @@ variable "proximity_placement_group_id" {
...
+variable "community_eviction_policy" {
+  description = "(Optional) The Eviction Policy which should be used for Virtual Machines within the Virtual Machine Scale Set powering this Node Pool. Possible values are Deallocate and Delete. Changing this forces a new resource to be created."
+  type        = string
+  default     = "Delete"
+}
+variable "community_spot_max_price" {
+  description = "(Optional) The maximum price you're willing to pay in USD per Virtual Machine. Valid values are -1 (the current on-demand price for a Virtual Machine) or a positive value with up to five decimal places. Changing this forces a new resource to be created."
+  type        = string
+  default     = "-1"
+}
```
saschjmil, line 133: Can these variables be used independently? Can someone set a max price on regular priority nodes? Just curious.

saschjmil (commented):
This looks really good. I have a few questions.

variables.tf
Carus11: - Regular is the default value when not provided in the code, so it defaults to that value here. 
- Delete should be the default behaviour 
- The -1 and no specific max price means the max price is equal to the on-demand price of the instance

I've tested with my tfvars enabling spot node pools, and I also ran a test with the sample-input-defaults.tfvars file.

When running with the sample-input-defaults.tfvars file I observed the following in terraform plan:

```
      + priority                = "Regular"
      + scale_down_mode         = "Delete"
      + spot_max_price          = -1
```

The priority = Regular is the gate to the spot functionality, and in this case means it will not be a spot node pool.

modules/aks_node_pool/variables.tf
Carus11: Testing this out in a environment, I've set the CAS node pool to Regular with a max price of 3.50
Terraform plan does not complain, it will accept it initially. 
When you apply there are some validation errors which will stop you, 

```
Error: `eviction_policy` can only be set when `priority` is set to `Spot`
[And also]
Error: `spot_max_price` can only be set when `priority` is set to `Spot`
```

variables.tf
Carus11: When applying a tfvars without specifying any of these community attributes in a node pool block, it creates a node pool with the following properties observed in azure portal

image

docs/community/community_config_vars.md
Carus11: I've updated the file

saschjmil: Looks good.

variables.tf
saschjmil: Sounds good. Thanks for the clarification and the extra testing.

modules/aks_node_pool/variables.tf
Carus11: Ah yeah thats a good idea

Carus11: Ok it got a bit messy, given I dont want it to try to validate them if they are simply not specified in the tfvars file. But it should still handle the example tfvars and then properly validate the vars when you start using them.

saschjmil: The validation checks look good to me. I did notice in my testing that lower case values cause issues. It works with Regular, Spot, Delete, and Deallocate, but would fail on regular, spot, delete, and deallocate. I pushed a change to your branch to remove the `lower()` portions of the validation checks.

The testing infrastructure to handle these community features is still in progress. I don't think we should delay this PR until the testing infrastructure is ready.

saschjmil (approved):
Looks good to me. Thanks for your patience and flexibility!

ndickens (approved):

PR #465, opened 2025-03-26
ci: (PSKD-1515) adding release workflows

saschjmil:
Adding three new workflows to the project to help with organization and releases:

**PR Conventional Commit Validation** 
This workflow will require all PRs to have a valid conventional commit prefix. It assigns a label based on that prefix.

**Release Drafter**
This workflow will draft up the release notes based on the PRs that have been merged since the last release. It uses the PR labels to categorize the PRs.

**Update Semver**
This workflow will update the previous semantic versioned tags. For example, if we release v1.2.3, it will update v1.2, v1, and latest.

(Ignore the 50+ test commits. They will be squashed into one.)


ndickens (dismissed):

.github/workflows/convention-commits.yaml
ephesused, line 1: Would you mind renaming this to `conventional-commits.yaml`?

saschjmil: Thanks for catching that. I'll rename it.

ephesused (dismissed):

ephesused (approved):

dhoucgitter (approved):

Issue #450, opened 2025-03-13
MI created by IaC doesn't have permissions to manage BYO Network components


frozentank:
### Terraform Version Details

We are using BYO networking but allowing IaC to create and manage the MI for deploying AKS. This results in an error: CustomRouteTableMissingPermissions.

Looking into the terraform, I don't see any place that is adding permissions (at least Network Contributor) to the BYO networking components for the MI that's been created.

### Terraform Variable File Details

  "vnet_resource_group_name": "vnet-rg",
  "vnet_name":"vnet-name",
  "subnet_names":{
    "aks": "aks-subnet",
    "misc": "misc-subnet",
    "netapp": "netapp-subnet"
  },

### Steps to Reproduce

Have a subscription with a networking resource group, vnet, subnets, nsg and route tables. Attempt to deploy IaC while allowing it to create a MI.

### Expected Behavior

The cluster is provisioned. Permissions for the MI are present on the networking components to allow AKS to manipulate the route tables.

### Actual Behavior

Terraform crashes with the below error:
╷
│ Error: creating Kubernetes Cluster (Subscription: "my subscription"
│ Resource Group Name: "tjm-test-rg"
│ Kubernetes Cluster Name: "tjm-test-aks"): performing CreateOrUpdate: unexpected status 400 (400 Bad Request) with response: {
│   "code": "CustomRouteTableMissingPermission",
│   "details": null,
│   "message": "Managed identity or service principle must be given permission to read and write to custom route table id-of-the-route-table. Please see https://aka.ms/aks/customrt for more information",
│   "subcode": ""
│  }
│ 
│   with module.aks.azurerm_kubernetes_cluster.aks,
│   on modules/azure_aks/main.tf line 5, in resource "azurerm_kubernetes_cluster" "aks":
│    5: resource "azurerm_kubernetes_cluster" "aks" {
│ 

where id-of-the-route-table is the ID of AKS subnet's associated route table

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

frozentank:
Code change to resolve was merged into main.

PR #477, opened 2025-04-11
fix: Adding permission management for BYO networks #450

frozentank:
This will add the necessary permissions for a MI created by IaC to any BYO components. It still requires that the SP used has access to manage permissions on the BYO components. #450


modules/azurerm_vnet/main.tf
```diff
@@ -58,3 +58,16 @@ resource "azurerm_subnet" "subnet" {
...
+resource "azurerm_role_assignment" "existing_network_assignment" {
+    count = length(var.existing_subnets) == 0 ? 0 : (var.add_uai_permissions ? length(var.roles) : 0)
+    scope = data.azurerm_subnet.subnet["aks"].route_table_id
+    role_definition_name = var.roles[count.index]
+    principal_id = var.aks_uai_principal_id
+}
+
+resource "azurerm_role_assignment" "existing_vnet_assignment" {
+  count = var.name == null ? (var.add_uai_permissions ? length(var.roles) : 0) : 0
+  scope = data.azurerm_virtual_network.vnet[0].id
```
saschjmil, line 70: I'm getting an error here when I try to run without bringing my own network. Perhaps it needs to check to see if the tuple exists before indexing it.

│ Error: Invalid index
│
│   on modules/azurerm_vnet/main.tf line 70, in resource "azurerm_role_assignment" "existing_vnet_assignment":
│   70:   scope = data.azurerm_virtual_network.vnet[0].id
│     ├────────────────
│     │ data.azurerm_virtual_network.vnet is empty tuple
│
│ The given key does not identify an element in this collection value: the collection has no elements.

saschjmil (changes requested):
This looks good. I'm able to run through your scenario with the BYO network. Thanks for updating the doc as well!

However, if I don't specify a network, it fails. I've marked the line that is causing the failures.

modules/azurerm_vnet/main.tf
frozentank: Do you mind sharing what you are/are not setting?

saschjmil: I've included my tfvars below.

```
# ****************  REQUIRED VARIABLES  ****************
# These required variables' values MUST be provided by the User
prefix   = "chjmil10" 
location = "eastus2" 
# ****************  REQUIRED VARIABLES  ****************

# **************  RECOMMENDED  VARIABLES  ***************
default_public_access_cidrs = ["****/16", "****/32"] 
ssh_public_key              = "/path/to/id_rsa.pub"
# **************  RECOMMENDED  VARIABLES  ***************

# Tags can be specified matching your tagging strategy.
tags = {"resourceowner": "****@sas.com"} 
```

I ran another plan today with this tfvars and got the same error as before.

saschjmil:
Hey @frozentank, sorry for closing and re-opening your PR. I deleted the staging branch, which closed all PR's targeting it. I've updated your PR to point to the main branch.

frozentank:
@saschjmil Please try the test case again with these new changes. Hopefully, they are addressed with that bit flip. As far as I could test it seemed to be working but I can't deploy a cluster that large (all of my testing has been with smaller node pools)

saschjmil (approved):
Looks good to me. I was able to create a cluster with and without a BYO network setup. The clusters came up successfully.

PhanikumarThodupunoori (approved):
LGTM

Issue #482, opened 2025-04-15
Netapp Pool and Volumes bound to same size


frozentank:
### Is your feature request related to a problem? Please describe.

We have need to create multiple volumes within the same Netapp Pool. Right now, the size parameter is used to declare how large both of them are created. This means that the entire pool is consumed by the single volume and any solution to get another volume into that pool is quite ugly (resize the pool, overwrite the netapp module).

### Describe the solution you'd like

In a perfect world, we could describe a dynamic number of volumes to be declared into the Netapp Pool.

But outside of that, at least giving an optional variable for the size of the volume and just default to the pool size if that value is missing.

### Describe alternatives you've considered

Right now, I can fix this by overwriting the entire netapp module (too bad Terraform doesn't support overriding modules). This causes me to have to maintain a new module that keeps feature parity with the current module.

Alternatively, we could click-ops the pool into a larger size and then add the volume after the fact. This nullifies the statefile and prevents Terraform destroy. Well, it doesn't matter as much now since AzureRM won't destroy Netapp Volumes but maybe they will fix that back soon.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

saschjmil:
Hey @frozentank, thanks for the enhancement request. The IAC maintainers are not planning to pursue this request. However, you or the community are welcome to open up a PR with this enhancement.

I'm going to add a "Help Wanted" label to signal that this will be handled by the community.

frozentank:
Closing as the feature was merged into main.

PR #493, opened 2025-05-05
feat: Adding community option for Azure NetappFiles Volume size control.

frozentank:
Added the option `contrib_netapp_volume_size` a number optional field that will specify an alternative size of the netapp volume created when HA with Netapp Volumes are specified. This is to address #482.

The default value for the field is 0 which when this is specified the volume size will match the pool size which is the current implementation.

### DIFF
--- a/docs/community/community_config_vars.md
+++ b/docs/community/community_config_vars.md
@@ -20,5 +20,5 @@ Here is a table with the variables you would use to configure them
 
 | Name | Description | Type | Default | Release Added | Notes |
 | :--- | ---: | ---: | ---: | ---: | ---: |
-| contrib_enable_spot_nodes | Enable spot nodes | bool | false | 10.3.0 | |
-| contrib_netapp_volume_size | Size of the netapp volume | number | 0 | | Zero will disable, must be smaller than the Netapp Pool. The value is given in GB |
+| community_enable_spot_nodes | Enable spot nodes | bool | false | 10.3.0 | |
+| community_netapp_volume_size | Size of the netapp volume | number | 0 | | Zero will disable, must be smaller than the Netapp Pool. The value is given in GB |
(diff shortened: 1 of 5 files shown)


saschjmil:
Hey @frozentank, this looks really good. Could you update the variable name to use the `community_` prefix instead of `contrib_`? Sorry about the confusion, we switched our approach as we worked through #451.

frozentank:
Alright hopefully that will resolve the change. @saschjmil

docs/community/community_config_vars.md
```diff
@@ -23,7 +23,8 @@ To enable a Spot node pool in your AKS cluster using this module, configure the
 
 | Name | Description | Type | Default | Release Added | Notes |
 | :--- | ---: | ---: | ---: | ---: | ---: |
+| community_enable_spot_nodes | Enable spot nodes | bool | false | 10.3.0 | |
```
saschjmil, line 26:
```suggestion
```
This line was just filler while we figured out the structure. It doesn't actually do anything so it can be removed.

docs/community/community_config_vars.md
```diff
@@ -23,7 +23,8 @@ To enable a Spot node pool in your AKS cluster using this module, configure the
...
+| community_netapp_volume_size | Size of the netapp volume | number | 0 | | Zero will disable, must be smaller than the Netapp Pool. The value is given in GB |
```
saschjmil, line 27: Could you move this into its own section? It doesn't relate to spot nodes.

Ideally, there would be a new section that contained the community-driven features related to NetApp. 
* There would be a new entry in the Table of Contents. 
* The section would contain any warnings about the feature (if applicable).
* The section would use the same table columns as the existing spot node table

saschjmil (commented):
Thanks for updating the variable names. One last request and then it should be good.

frozentank:
@saschjmil How do those changes look?

saschjmil:
That looks good to me. Once you remove the original lines (lines 27 and 28), this will be ready to go. Thanks!

I'm also going to add the `feat` conventional commit to your PR title.

saschjmil (approved):
Looks good. Thanks for the contribution!

ndickens (approved):

PR #504, opened 2025-05-16
docs: Updating CONFIG-VARS ssh_public_key

frozentank:
Updating the CONFIG-VARS.md to reflect the usage of the ssh_public_key variable


saschjmil (approved):
Looks good to me

dhoucgitter:
Hi @frozentank, I think your addition is accurate but I'm wondering if folks might expect that opening an ssh session to the AKS nodes VMs is available with the default cluster SG and rules configuration. 
Those only allow ssh to jump and nfs VMs when present.

frozentank:
Thanks @dhoucgitter, you can still get into the nodes by running a pod within the cluster and then using SSH to any of the hosts. This is by default without needing to modify any rules. 

We actually want to be able to control the SSH key being set on the AKS nodes as we've found plenty of hosting issues by being able to connect into the nodes directly.

dhoucgitter (approved):
LGTM, thanks for the doc contribution.

PR #513, opened 2025-06-06
fix: change aks_pod_cidr based on aks_network_plugin (PSKD-1458)

saschjmil:
Update the aks_pod_cidr based on the value of the aks_network_plugin. Currently, if it is set to azure, it will use the wrong aks_pod_cidr which prevents SAS Studio from executing code.

Testing Scenarios:
| Scenario| Provider| K8s version | Deploy Method | aks_network_plugin | Notes |
|--------|--------|--------|--------|--------|---------|
| 1| azure | 1.31.8 | Local tooling | azure | Successful Viya deployment. I was able to execute SAS code in Studio. |
| 2 | azure| 1.31.8 | Local tooling | kubenet | Successful Viya deployment. I was able to execute SAS code in Studio. |

### DIFF
--- a/outputs.tf
+++ b/outputs.tf
@@ -148,3 +148,7 @@ output "cluster_node_pool_mode" {
 output "cluster_api_mode" {
   value = var.cluster_api_mode
 }
+
+output "aks_network_plugin" {
+  value = var.aks_network_plugin
+}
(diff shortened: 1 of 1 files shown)


dhoucgitter (dismissed):

PhanikumarThodupunoori (dismissed):

PhanikumarThodupunoori (approved):
LGTM

ndickens (approved):

PR #519, opened 2025-06-19
fix: (518) allow aks_pod_cidr to be set when cni is azure w/ overlay

ajeffowens:


saschjmil (approved):
Looks good to me. Thanks for the contribution!

dhoucgitter (approved):
LGTM also.

PR #594, opened 2026-04-20
fix: allow null values for NetApp availability and replication zones

ajeffowens:
Closes #580 

# Summary
This PR fixes Terraform validation failures when NetApp zone inputs are set to null.  `netapp_availability_zone` can now safely be set to `null` to trigger a non-zonal deployment.  

Previously, validation used an OR expression with 'contains':

```
var.netapp_availability_zone == null || contains([...], var.netapp_availability_zone)
```

A `null` value would fail with:

```
Invalid value for "value" parameter: argument must not be null.
```

Because 'contains' does not accept null as its value argument.

# What Changed
The validation logic is now:

```
var.netapp_availability_zone == null ? true : contains([...], var.netapp_availability_zone)
```

The conditional operator x ? y : z is the reliable lazy form in Terraform for this case, because only one branch is evaluated.
So the revised expression is the safer pattern for nullable inputs.

* `null` is now accepted safely for nullable zone variables without triggering contains(..., null).
* Valid non-null values remain unchanged: 1, 2, 3.
* Invalid non-null values still fail validation as expected.


ajeffowens:
Note that I also updated `netapp_replication_zone` var in this PR since it was previously claiming to allow a null value, though it too was using the same invalid expression.  I am not sure whether or not it makes sense for this var to ever be null though...

Another aside, that is not addressed in this PR, is that I think the default value for `netapp_availability_zone` should probably be a local that is calculated to the same value as `node_pools_availability_zone`.

Carus11:
This is focused and directly addresses the issue. Looks good to me, the comment about the netapp availability zone should calculation should indeed be added to the backlog.

